PR #1, opened 2022-01-18
chore: Import base from merino-admin

mythmon:
I think this is ready to merge as is, assuming review passes, but it has a ways
to go before it is production ready.  This is mostly a way to provide a base we
can all work on.

This relies on a branch I've made on the Merino repository. The linkage between
this repository and Merino is definitely going to change, at least a little
bit, as things progress. Also we'll want to add Contile specific bits as well.


README.md
```diff
@@ -1,2 +1,27 @@
...
+
+This is a tool to manage dynamic settings for Contextual Services projects like
+Merino and Contile. The settings managed here can be changed at runtime,
+reducing the need for deployments of the services that use it.
+
+To use consvc-shepherd, you'll need a Python development environment, and
+[Poetry][]. With that ready, you can set up the Django site:
+
+```shell
+$ poetry shell
```
jrconlin, line 11: might want to update this since we're not using poetry.

jrconlin: Also, do we need to run `collectstatic` here at all? I've had installs fail to find these files in the past.

jrconlin (changes requested):

README.md
mythmon: For development, Django automatically routes simple static files through. So if the following stays true, we're fine

1. Not doing anything fancy with static files (like preprocessors)
2. Running with `./manage.py runserver`
3. Using `DEBUG=true`

Of course, 2 and 3 will change in production, but the quick start docs don't need to worry about that.

mythmon: That being said, I realize now that this quick start doesn't mention `DEBUG=true` at all. I'll add some more docs about environment management to deal with that.

jrconlin (approved):

PR #3, opened 2022-03-03
stubbed out models and admin

tiftran:
Thought it would be best that I throw something up now

- roughly stubbed out models and the admin
- "Advertisers" are not currently connected to "SettingsSnapshot"
-  I've placed contile related models in `contile.py` they would probably move into contile later like merino?


consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
```diff
@@ -0,0 +1,50 @@
...
+                ('created_by', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL)),
+                ('launched_by', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='launched_by', to=settings.AUTH_USER_MODEL)),
+                ('json_settings', models.JSONField(blank=True, null=True)),
+            ],
+        ),
+        migrations.CreateModel(
+            name='AdvertiserUrl',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('domain', models.URLField()),
```
jrconlin, line 42: This might be hard to do, but I think it might be useful to construct the `advertiser_url` as a group. 
Advertisers would have one url per geo, and that URL data should match the geo specified. (so we avoid the "sending example.fr to DE because it's listed as a valid URL" problem. )

e.g.
```json
"advertiser_url": {
    geo: {
        "domain": domain,
        "path": path,
        "exact": bool,
    },
    ...
}
```

consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
```diff
@@ -0,0 +1,50 @@
...
+                ('advertiser', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='ad_urls', to='consvc_shepherd.advertiser')),
+                ('exact', models.BooleanField(default=False)),
+                ('path', models.CharField(default=False, max_length=128)),
+                ('prefix', models.BooleanField(default=False)),
```
jrconlin, line 46: Thinking about this, right now we have a boolean of paths being either "exact" or "prefix". We'll never have both. We should either drop one of these, or maybe have a "path_type" that's an enumeration if we ever decide we want to have more than two types. (I don't think we will, but I am also terrible at predicting the future.)

consvc_shepherd/models/contile.py
```diff
@@ -0,0 +1,33 @@
...
+                                   null=False,
+                                   related_name="ad_urls",
+                                   on_delete=models.CASCADE, )
+
+    # TODO might want to consider whether we want a subset of countries
+    geo = CountryField()
+    domain = models.URLField()
+    path = models.CharField(max_length=128)
+    exact = models.BooleanField(default=False)
+    prefix = models.BooleanField(default=False)
```
jrconlin, line 25: See the comment in the migrations field, but I think we can probably drop one of these, since it's a boolean state and we're not going to have both set to True.

consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
```diff
@@ -0,0 +1,50 @@
...
+        ),
+        migrations.CreateModel(
+            name='SettingsSnapshot',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(blank=True, max_length=5)),
+                ('created_on', models.DateTimeField(auto_now_add=True)),
+                ('created_by', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL)),
+                ('launched_by', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='launched_by', to=settings.AUTH_USER_MODEL)),
+                ('json_settings', models.JSONField(blank=True, null=True)),
```
jrconlin, line 35: Don't we want to record modification info too? (date, userid)

consvc_shepherd/storage.py
```diff
@@ -0,0 +1,14 @@
+import logging
+
+from django.core.files.storage import default_storage
+from django.conf import settings
+
+
+def send_to_storage(content_name, content):
```
jrconlin, line 7: I'm guessing a TODO here is to add the modification data.

consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
tiftran: I went back and forth with this, figured that having two boolean fields would be a quick and easy way to get two radio buttons in the UI without doing anything. Took another look, it wasn't so hard to do that with one boolean field after all :)

consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
tiftran: I think this is achievable at the point where we convert the models into json, moving geo into another "level" would mean another level of model nesting.

tiftran: that being said i realize that `path` should actually be `paths` plural. Would it be fair to say that 
```
"advertiser_url": {
    geo: {
        "domain": domain,
        "paths": [],
        "exact": bool,
    },
    ...
}
```

is not what we want? and the fact that for a given domain in a geo paths could be a mix of exact and prefix?

consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
tiftran: `created_on`,`created_by`,`launched_by` should achieve this right? but a `launch_date` would probably be good too

consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
jrconlin: yes, you're right. `paths` should be a list of dicts that contain the `path` and `exact` bool.

consvc_shepherd/migrations/0001_squashed_0004_advertiserurl_geo.py
jrconlin: Ah, I hadn't considered that these would iterate by snapshot. If that's the case, then yes, the metadata would be relevant per snapshot.

jrconlin (approved):
There are a few extra tweaks, but they can be covered in the next PR.

PR #4, opened 2022-03-10
add docker-compose file

tiftran:
added a docker-compose file,  for people (and myself) who want to work with a docker work-flow

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,45 @@
+version: "3"
+
+services:
+  db:
+    restart: always
+    image: postgres:14.0
+    volumes:
+      - db_volume:/var/lib/postgresql
+    environment:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      POSTGRES_HOST_AUTH_METHOD: trust
+    ports:
+      - "5432:5432"
+    networks:
+      - private_nw
+
+  app:
+    tty: true
+    env_file:
+      - .env
```
jrconlin, line 20: Is this file missing?

tiftran: ah yes i should probably have a .env.sample

pjenvey (approved):
Maybe an addition to the README about setting up an .env too?

PR #13, opened 2022-03-14
feat: add fields click_hosts, impression_hosts, and position to adurl model

### LINKED ISSUE #12: Add additional fields to AdUrls model
The following fields need to be added:
`click_hosts` - ArrayField of Strings that is optional
`impression_hosts` - ArrayField of Strings that is optional
`position` - integer field

tiftran:
- added fields to ad url model using Array Field 
- made adjustments to admin to accommodate new fields
- switched ad_url to stacked inline because tabular inline started looking too wide
- since Array field is introduced and is postgres specific, the database field has been changed to no longer use sqllite as a default

closes #12


consvc_shepherd/models/contile.py
```diff
@@ -12,19 +18,32 @@ def to_dict(self):
...
+        on_delete=models.CASCADE,
+    )
 
     # TODO might want to consider whether we want a subset of countries
     geo = CountryField()
     domain = models.URLField()
     path = models.CharField(max_length=128)
     matching = models.BooleanField(choices=MATCHING_CHOICES, default=False)
-
+    click_hosts = ArrayField(
```
ncloudioj, line 36: It's worth noting that both `click_hosts` and `impression_hosts` are shared by all advertisers for the same partner (e.g. adM) and that applies both to Contile and Merino. Perhaps we should introduce the `Partner` concept to the model so we don't need to set those for every advertiser?

Also, both of them could be optional, some partners (e.g. Wikipedia) do not require us to report impressions and clicks back to them.

tiftran: So I've been looking in between [adm_settings.json](https://github.com/mozilla-services/contile/blob/main/adm_settings_test.json),
[adm_settings.json](https://github.com/mozilla-services/contile/blob/main/integration-tests/volumes/contile/adm_settings.json), [values.yaml](https://github.com/mozilla-services/cloudops-infra/blob/7a9183f1d8ea78d96766cc2615fa9c1abf7ffeb0/projects/topsites/k8s/charts/contile/values.yaml)
they are all slightly different, would it be safe to assume the yaml file is the best source of truth of what we ultimately want? in this case, we don't need `impression_hosts` and `click_hosts` as a field per advertiser? and they're set a the `DEFAULT` level?

ncloudioj: Yeah, the YAML file is what we're using in production.

> we don't need impression_hosts and click_hosts as a field per advertiser?

Correct, we don't need them at the advertiser level.

Also, I think the current setting structure defined in YAML can be treated as a base for the model that we're looking to build, so feel free to modify & augment it if necessary.

tiftran: So I added a `Partner` model and an abstract `AdUrl` model which is inherited by `PartnerAdurl` and `AdvertiserUrl`

consvc_shepherd/migrations/0001_initial.py
```diff
@@ -0,0 +1,182 @@
+# Generated by Django 4.0.2 on 2022-03-16 07:13
```
tiftran, line 1: I had to squash/recreate the migrations, after i got myself into trouble with some weird circular model thing...

consvc_shepherd/models/contile.py
```diff
@@ -7,8 +9,50 @@
...
+        for hostname in hostname_list:
+            if "." not in hostname or any(
+                [not h.isalnum() for h in hostname.split(".")]
+            ):
+                raise ValidationError(
+                    f"{hostname} is not a valid hostname, hostnames should only contain alpha numeric characters and '.'"
+                )
+
+    def clean(self):
+        super().clean()
```
ncloudioj, line 35: Lol, so our partners will be "super clean".

consvc_shepherd/migrations/0001_initial.py
```diff
@@ -0,0 +1,182 @@
...
+                        null=True,
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name="launched_by",
+                        to=settings.AUTH_USER_MODEL,
+                    ),
+                ),
+            ],
+        ),
+        migrations.CreateModel(
+            name="PartnerAdUrl",
```
ncloudioj, line 105: Nit: I wonder if we can squash/recreate the migrations again here so that `PartnerAdUrl` never shows up in here and the subsequent [0003_delete_partneradurl.py](https://github.com/mozilla-services/consvc-shepherd/pull/13/files#diff-393ced2abff12a059997f30ee24b07437d5a9e39d45b0b4517f0c136fd40e346). Please disregard this if that makes everything messy.

tiftran: yup, makes sense, just want to check if there were any further changes before squashing again

consvc_shepherd/tests/test_models_contile.py
```diff
@@ -0,0 +1,22 @@
+from django.core.exceptions import ValidationError
+from django.test import TestCase
+
+from consvc_shepherd.models import Partner
+
+
+class TestPartnerModel(TestCase):
```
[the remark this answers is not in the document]
ncloudioj: Doh, just realized that we've switched over to PostgreSQL from SQLite. We have to set up Postgres in CI to make this work. Should be straightforward to do with this [doc](https://circleci.com/docs/2.0/databases/).

ncloudioj (approved):

consvc_shepherd/migrations/0001_initial.py
tiftran: TIL you can't squash, squashed migrations. So I tossed out the old ones and made a new one

consvc_shepherd/tests/test_models_contile.py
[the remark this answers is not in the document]
tiftran: Leaving this out of the PR, will file another one for this

PR #18, opened 2022-03-22
[CONSVC-1346]feat: validation for advertiser urls

### LINKED ISSUE #9: Add validation to Advertiser/AdUrls
- Validation should go in `clean` methods
- Rules for validation can be found [here](https://docs.google.com/document/d/1Ck8WiXavFnEvvF0dOzPlBlLTaaFffI5wBe_BTtXNL5k/edit#heading=h.a8t46dk2j799)

tiftran:
closes #9


consvc_shepherd/models/contile.py
```diff
@@ -104,14 +107,31 @@ class AdvertiserUrl(models.Model):
...
     def clean(self) -> None:
-        pass
+        self.validate_unique()
+        is_valid_host(self.domain)
+        if not self.path.startswith("/"):
+            raise ValidationError(INVALID_PATH_ERROR)
 
-    def __str__(self):
-        return f"{self.geo}{self.domain}{self.path}"
+        if self.get_matching_display() == "prefix" and self.path.endswith("/"):
```
ncloudioj, line 124: I think we always require the prefix paths to end with "/" in [Contile](https://github.com/mozilla-services/contile/blob/main/src/adm/settings.rs#L58-L59).

consvc_shepherd/models/contile.py
```diff
@@ -104,14 +107,31 @@ class AdvertiserUrl(models.Model):
     )
 
     geo = CountryField()
-    domain = models.URLField()
+    domain = models.CharField(max_length=80)
```
ncloudioj, line 110: Nit: per [Wikipedia](https://en.wikipedia.org/wiki/Hostname), the length of hostname could go up to 253 ASCII characters.

consvc_shepherd/models/contile.py
```diff
@@ -104,14 +107,31 @@ class AdvertiserUrl(models.Model):
...
+            raise ValidationError(INVALID_PREFIX_PATH_ERROR)
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        return super(AdvertiserUrl, self).save(*args, **kwargs)
+
+
+def is_valid_host(host):
+
+    if not all([h.isalnum() or h in [".", "-"] for h in host]) or "." not in host:
```
ncloudioj, line 134: I think we can also check the total number of dots in the hostname. Should be either 1 or 2 dots to cover both `{leaf-domain}.{second-level-domain}.{top-domain}` and `{second-level-domain}.{top-domain}`.

consvc_shepherd/models/contile.py
ncloudioj: Let's also update all the unit tests below to reflect that.

ncloudioj (changes requested):

consvc_shepherd/models/contile.py
jrconlin: IIRC, [RFC 2181](https://datatracker.ietf.org/doc/html/rfc2181#section-11) says that a max Fully Qualified Domain Name length of 255 octets, including the full stops `.` (one octet = one traditional 8 bit byte character). This should also cover IDNA entries as well, if we ever get those.

consvc_shepherd/models/contile.py
```diff
@@ -104,14 +107,34 @@ class AdvertiserUrl(models.Model):
...
+
+def is_valid_host(host):
+
+    if not all([h.isalnum() or h in [".", "-"] for h in host]):
+        raise ValidationError(
+            f"{host}: hostnames should only contain alpha numeric characters '-' and '.'"
+        )
+    if not 2 <= len(host.split(".")) <= 3 or "" in host.split("."):
+        raise ValidationError(
+            f"{host}: hostnames should have the structure <leaf-domain>.<second-level-domain>.<top-domain> and <second-level-domain>.<top-domain>"
```
ncloudioj, line 139:
```suggestion
            f"{host}: hostnames should have the structure <leaf-domain>.<second-level-domain>.<top-domain> or <second-level-domain>.<top-domain>"
```

consvc_shepherd/migrations/0003_alter_advertiserurl_domain.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("consvc_shepherd", "0002_alter_advertiserurl_domain"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="advertiserurl",
+            name="domain",
+            field=models.CharField(max_length=253),
```
ncloudioj, line 16: As @jrconlin pointed out, let's do 255 (253 + 2 dots) here :)

consvc_shepherd/models/contile.py
```diff
@@ -104,14 +107,34 @@ class AdvertiserUrl(models.Model):
     )
 
     geo = CountryField()
-    domain = models.URLField()
+    domain = models.CharField(max_length=253)
```
ncloudioj, line 110: As @jrconlin pointed out, let's do 255 (253 + 2 dots) here :)

consvc_shepherd/tests/test_models_contile.py
```diff
@@ -74,31 +82,133 @@ def test_to_dict_produces_correctly(self):
...
+                geo="France",
+                domain="example.com",
+                path="/prefix/",
+                matching=False,
+                advertiser=self.advertiser,
+            ).count(),
+            1,
+        )
+
+    def test_ad_url_with_valid_exact_saves_correctly(self):
```
ncloudioj, line 197: For completeness, could you add another test case to ensure FQDNs (e.g. `www.example.com`) are accepted?

ncloudioj (approved):
r+ with a few more nits.

Looks great, thanks!

consvc_shepherd/models/contile.py
tiftran: eek! didnt see @jrconlin's comment, yes will do!

consvc_shepherd/models/contile.py
tiftran: ugh... english is hard 😅

PR #36, opened 2022-04-22
[CONSVC-1733] chore: added tests to ensure coverage

tiftran:
Added tests for coverage, this should techincally put line coverage at 95%

closes #24

PR #37, opened 2022-04-22
[CONSVC-1709] chore: enable tests and coverage for circle

### LINKED ISSUE #15: Set up circle to run tests
Currently the circleci config is set so that there is a migration check and lining.

We would want it to check for tests passing too.
This will require docker user and password env vars? and introducing postgres into the config

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1709)

tiftran:
Enables tests for circle.
Docker credentials are temp, since it's current using mine

closes #15

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,15 @@ jobs:
     # Change the version below to your required version of python
     docker:
       - image: cimg/python:3.10.2
+        environment:
+          DJANGO_SETTINGS_MODULE: consvc_shepherd.settings
+      - image: cimg/postgres:14.2
+        environment:
+         POSTGRES_HOST_AUTH_METHOD: trust
+        auth:
+          username: $DOCKER_USER_TEMP
+          password: $DOCKER_PASS_TEMP
+
     # Checkout the code as the first step. This is a dedicated CircleCI step.
     # The python orb's install-packages step will install the dependencies from a Pipfile via Pipenv by default.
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -33,10 +42,18 @@ jobs:
...
+            if [ "${DOCKER_USER_TEMP}" == "" ] || [ "${DOCKER_PASS_TEMP}" == "" ]; then
+              echo "Skipping Login to Dockerhub, credentials not available."
+            else
+              echo "${DOCKER_PASS_TEMP}" | docker login -u="${DOCKER_USER_TEMP}" --password-stdin
+            fi
+      - run:
+          name: Run tests
           # This assumes pytest is installed via the install-package step above
-      #    command: pytest
+          command: pytest --cov --cov-report=term-missing --cov-fail-under=95
```
ncloudioj, line 56: How about `--cov-fail-under=100`? ;)

ddurst: 😍

ddurst: (I am kidding)

tiftran: 😆

tiftran: 95 wasn't arbitrary 🙂. With #36 it's at 95-ish%, lines that weren't covered were the `__str__` methods, which I decided probs weren't worth the trouble

tiftran:
merging with $DOCKER_USER_TEMP and $DOCKER_PASS_TEMP. Will create issue to use real credentials once they're set

PR #39, opened 2022-04-28
[CONSVC-1729] feat: introduce approval for partners

### LINKED ISSUE #23: Add method to enforce two person sign off in partner model
Recommended in the RRA, we would like to enforce dual sign off of partner settings before they are used

Introducing an "is_active" (there probably is a better name for it) field to the Partner Model, that is read-only in the change form.
The "is_active" field will always flip to false everytime the model is updated.
Adding an admin action that switches "is_active" to true.

Considerations:
What happens when a user wants to create a snapshot when the Partner Model has "is_active" false?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1729)

tiftran:
decided that if a partner setting has changed, you are unable to create a snapshot with that partner unless it has been approved. 
closes #23


consvc_shepherd/models.py
```diff
@@ -23,3 +24,14 @@ class SettingsSnapshot(models.Model):
 
     def __str__(self):
         return f"{self.name}: {self.created_on}"
+
+    def clean(self):
+        if (
+            self.settings_type.is_active is False
```
ncloudioj, line 30: Nit:

```suggestion
            !self.settings_type.is_active
```

contile/admin.py
```diff
@@ -12,6 +12,23 @@ class Meta:
...
+    else:
+        partner = queryset[0]
+        if partner.last_updated_by != request.user:
+            partner.is_active = True
+            partner.last_approved_by = request.user
+            partner.save()
+            messages.info(request, f"Partner: {partner.name} has been approved")
+        else:
+            messages.error(
+                request, "Approver is the same user that updated the partner"
```
ncloudioj, line 28: Nit: let's also point out how this should be done, such as "This change can't be approved by the same editor, please get another reviewer for approval".

consvc_shepherd/models.py
tiftran: `! self.settings_type.is_active` isn't valid python will be changing it to `not self.settings_type.is_active`

ncloudioj: Doh, my Python is getting Rusty 😂.

ncloudioj (approved):
LGTM, thanks!

PR #41, opened 2022-05-16
[CONSVC-1739] "exact" path matching as default

### LINKED ISSUE #32: Make "exact" path matching as default in the advertiser form
Since "exact" matching is much more common than "prefix" matching in practice, let's make it the default of that radio button.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1739)

tiftran:
closes #32


ncloudioj (changes requested):
Looks like this requires a migration.

tiftran:
gah, you're right, forgot to include it... lemme do that

ncloudioj (approved):
🚢

PR #42, opened 2022-05-16
[CONSVC-1736] Disable ability to delete Settings Snapshots

### LINKED ISSUE #29: Disable ability to delete Settings Snapshots
Setting Snapshots should be preserved as logs. As a result we don't want Setting Snapshots to be able to be deleted in
django admin.

tiftran:
Snapshots shouldn't be able to be deleted if they were launched 
closes #29


tiftran:
thanks @ncloudioj

PR #49, opened 2022-08-19
chore: dockerflow setup

tiftran:


consvc_shepherd/settings.py
```diff
@@ -105,3 +107,23 @@
...
 GS_BUCKET_NAME = env("GS_BUCKET_NAME", default="")
+
+LOGGING = {
+    "version": 1,
+    "formatters": {
+        "json": {"()": "dockerflow.logging.JsonLogFormatter", "logger_name": "shepherd"}
+    },
+    "handlers": {
+        "console": {
+            "level": "DEBUG",
```
ncloudioj, line 118: Using `DEBUG` in production could be quite noisy, would you consider using different levels in different environments?

requirements.txt
```diff
@@ -41,6 +41,8 @@ django-stubs==1.10.1
     # via -r requirements.in
 django-stubs-ext==0.4.0
     # via django-stubs
+dockerflow==2022.1.0
```
ncloudioj, line 44: There is a newer version for dockerflow.

ncloudioj (approved):
r+ with comments.

PR #50, opened 2022-09-13
Add .circleci/config.yml

tiftran:


.circleci/config.yml
```diff
@@ -67,10 +77,52 @@ jobs:
...
+          name: Restore Docker image cache
+          command: docker load -i /home/circleci/cache/docker.tar
+      - run:
+          name: Deploy to Dockerhub
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "main" ]; then
+              DOCKER_TAG="${CIRCLE_SHA1}"
+            fi
+            
+            if echo "${CIRCLE_BRANCH}" | grep '^feature\..*' > /dev/null; then
```
ncloudioj, line 97: Does this mean that we will deploy any of the following changes to dockerhub?

- Any changes merged into the `main` branch
- Any branches whose branch name starts with `feature\`

For the latter case, what happens when the feature branch gets merged to `main`? Will we publish the same image again with a different tag?

.circleci/config.yml
```diff
@@ -67,10 +77,52 @@ jobs:
...
+              DOCKER_TAG="${CIRCLE_BRANCH}"
+            fi
+            
+            if [ -n "${CIRCLE_TAG}" ]; then
+              DOCKER_TAG="$CIRCLE_TAG"
```
ncloudioj, line 102:
```suggestion
              DOCKER_TAG="${CIRCLE_TAG}"
```
Nit: to be consistent, let's always use parentheses for string interpolations. There are a few other cases down below.

.circleci/config.yml
ncloudioj: Looks like we also use this branch filter for the `deploy` job. I guess we can drop this check here since `${CIRCLE_BRANCH}` should always be `main`.
```
filters:
   branches:
        only: main
```

ncloudioj (approved):

PR #53, opened 2022-09-27
fix: use circleci base image for deploy

ncloudioj:


.circleci/config.yml
```diff
@@ -79,14 +84,16 @@ jobs:
 
   deploy:
     docker:
-      - image: docker:18.02.0-ce
+      - image: cimg/base:2022.08
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
     steps:
+      - attach_workspace:
```
tiftran, line 92: sooo i checked out this PR remove the `main` branch requirement for the deploy and i got [this](https://app.circleci.com/pipelines/github/mozilla-services/consvc-shepherd/161/workflows/6b5ea006-7594-41b5-9734-7409511f3a44/jobs/180). I think it requires a `setup_remote_docker` step before this.  But then after adding that circleci gives me [this](https://app.circleci.com/pipelines/github/mozilla-services/consvc-shepherd/162/workflows/8d3e7e2a-ef1a-4eab-ad17-4ed8f8333c8d/jobs/184). `denied: requested access to the resource is denied` for docker push. Is there a permission thing we need to enable?

ncloudioj: Good catch on `setup_remote_docker`. I will fix that in this PR.

I couldn't find ` mozilla/shepherd` on docker hub. Have we created that yet? If not, let's work with SRE on setting it up. Looks like environment variables (`DOCKER_USER, DOCKER_PASS, DOCKERHUB_REPO`) are already created on CircleCI. Let's also make sure that `DOCKERHUB_REPO` is set to `mozilla/shepherd` rather than `shepherd`.

tiftran: OHHHH it's https://hub.docker.com/r/mozilla/consvc-shepherd

tiftran (approved):
lgtm, jbuck has fixed the docker push issue, hopefully it pushes 🙂

PR #59, opened 2022-10-19
Naming convention for bucket files

tiftran:
Introduce an env for the bucket file name
The idea is that every time a new snapshot is pushed, the existing file would be rewritten under a different file name with the original creation date appended to the name.

Thoughts?

### DIFF
--- a/consvc_shepherd/storage.py
+++ b/consvc_shepherd/storage.py
@@ -6,16 +6,18 @@ from django.core.files.storage import default_storage
 
 def send_to_storage(content):
     if settings.DEBUG:
-        logging.info(f"Sending to storage, name:{settings.GS_BUCKET_FILE_NAME}, content: {content}")
+        logging.info(
+            f"Sending to storage, name:{settings.GS_BUCKET_FILE_NAME}, content: {content}"
+        )
     else:
         current_file_name = f"{settings.GS_BUCKET_FILE_NAME}-current.json"
         # write content in -current file to another file
         if default_storage.exists(current_file_name):
             creation_date = default_storage.get_created_time(current_file_name)
 
-            current_file = default_storage.open(current_file_name,"r")
+            current_file = default_storage.open(current_file_name, "r")
             current_file_contents = current_file.read()
-            previous_file_name =f"{settings.GS_BUCKET_FILE_NAME}-{creation_date}"
+            previous_file_name = f"{settings.GS_BUCKET_FILE_NAME}-{creation_date}"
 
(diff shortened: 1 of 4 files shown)


tiftran:
left this in draft....

consvc_shepherd/storage.py
```diff
@@ -4,10 +4,25 @@
...
-def send_to_storage(content_name, content):
+def send_to_storage(content):
     if settings.DEBUG:
-        logging.info(f"Sending to storage name:{content_name}, content: {content}")
+        logging.info(
+            f"Sending to storage, name:{settings.GS_BUCKET_FILE_NAME}, content: {content}"
+        )
     else:
-        file = default_storage.open(content_name, "w")
+        current_file_name = f"{settings.GS_BUCKET_FILE_NAME}-current.json"
```
ncloudioj, line 13: By convention, "-latest" is preferred over "-current".
Does Contile also read the file with `-current`, if it's not too bad to update that in Contile, we can use "-latest" here.

consvc_shepherd/storage.py
```diff
@@ -4,10 +4,25 @@
...
+        # write content in -current file to another file
+        if default_storage.exists(current_file_name):
+            creation_date = default_storage.get_created_time(current_file_name)
+
+            current_file = default_storage.open(current_file_name, "r")
```
ncloudioj, line 18: It's unfortunate that django-storage doesn't support renaming for GCS.
 
How about we create two files for each snapshot? One with the current datetime in the name, the other with "latest"? Then we don't have to do this "read-write-write" next time.

tiftran: much cleaner

consvc_shepherd/storage.py
```diff
@@ -2,12 +2,22 @@
...
-        logging.info(f"Sending to storage name:{content_name}, content: {content}")
+        logging.info(
+            f"Sending to storage, name:{settings.GS_BUCKET_FILE_NAME}, content: {content}"
+        )
     else:
-        file = default_storage.open(content_name, "w")
-        file.write(content)
-        file.close()
+        latest_file_name = f"{settings.GS_BUCKET_FILE_NAME}_latest.json"
+        date_file_name = f"{settings.GS_BUCKET_FILE_NAME}_{timezone.now()}"
```
ncloudioj, line 15:
```suggestion
        date_file_name = f"{settings.GS_BUCKET_FILE_NAME}_{timezone.now()}.json"
```

ncloudioj (approved):
r+ wc

PR #66, opened 2022-11-23
fix: remove header prefix for openidc

tiftran:
`X-Goog-Authenticated-User-Email` returns value of format `accounts.google.com:example@gmail.com`
So we need to remove that before going ahead and creating a user
ref: https://cloud.google.com/iap/docs/identity-howto#getting_the_users_identity_with_signed_headers

closes #CONSVC-2120


tiftran:
cc: @jbuck

ncloudioj (approved):

PR #67, opened 2022-11-23
fix: auth gcp jwt header[MERGE ON HOLD]

tiftran:
Using JWT is preferred over using simply the google email header.

I've made changes so that the email header should work when `debug=True`. It should also work if the email header isn't there.

JWT validation only occurs when `debug=False`

### DIFF
--- a/consvc_shepherd/settings.py
+++ b/consvc_shepherd/settings.py
@@ -21,6 +21,7 @@ DEBUG = env("DEBUG", default=False)
 DEV_USER_EMAIL = "dev@example.com"
 OPENIDC_HEADER = env("OPENIDC_HEADER", default=None)
 OPENIDC_HEADER_PREFIX = env("OPENIDC_HEADER_PREFIX", default=None)
+IAP_AUDIENCE = env("IAP_AUDIENCE", default=None)
 ALLOWED_HOSTS: List[str] = ["*"]
 
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
--- a/consvc_shepherd/settings.py
+++ b/consvc_shepherd/settings.py
@@ -130,6 +130,10 @@ LOGGING = {
         },
     },
     "loggers": {
+        "shepherd": {
+            "handlers": ["console"],
+            "level": env("SHEPHERD_ENV", default="DEBUG"),
+        },
         "request.summary": {
             "handlers": ["console"],
             "level": env("SHEPHERD_ENV", default="DEBUG"),
--- a/openidc/middleware.py
+++ b/openidc/middleware.py
@@ -7,6 +7,8 @@ from django.http import HttpResponse
 from google.auth.transport import requests
 from google.oauth2 import id_token
 
+logger = logging.getLogger("shepherd")
+
 
 def validate_iap_jwt(request):
     iap_jwt = request.META.get(settings.OPENIDC_HEADER)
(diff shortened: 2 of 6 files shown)


openidc/middleware.py
```diff
@@ -1,7 +1,31 @@
...
+
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.middleware import AuthenticationMiddleware
 from django.http import HttpResponse
+from google.auth.transport import requests
+from google.oauth2 import id_token
+
+
+def validate_iap_jwt(request):
```
tiftran, line 11: got this from https://cloud.google.com/iap/docs/signed-headers-howto#retrieving_the_user_identity

tiftran:
Ready for review, but I think put on hold for merge, considering it will cause the dev instance some complications where DEBUG=False there but is currently using the email header

openidc/middleware.py
```diff
@@ -1,7 +1,31 @@
...
+    iap_jwt = request.META.get(settings.OPENIDC_HEADER)
+    try:
+        decoded_jwt = id_token.verify_token(
+            iap_jwt,
+            requests.Request(),
+            audience=settings.IAP_AUDIENCE,
+            certs_url="https://www.gstatic.com/iap/verify/public_key",
+        )
+        return decoded_jwt["email"]
+    except ValueError as e:
```
ncloudioj, line 21: Per IAP's manual page, it uses `Exception`. Are you sure that `ValueError` would be sufficient?

consvc_shepherd/settings.py
```diff
@@ -21,6 +21,7 @@
 DEV_USER_EMAIL = "dev@example.com"
 OPENIDC_HEADER = env("OPENIDC_HEADER", default=None)
 OPENIDC_HEADER_PREFIX = env("OPENIDC_HEADER_PREFIX", default=None)
+IAP_AUDIENCE = env("IAP_AUDIENCE", default=None)
```
ncloudioj, line 24: Shall we add this to env.example as well?

openidc/middleware.py
```diff
@@ -1,7 +1,31 @@
...
+        logging.error(f"**ERROR: JWT validation error {e}**'.")
```
ncloudioj, line 22: This will use the ROOT logger which might not be in Mozlog format. You can add a new logger for this [here](https://github.com/mozilla-services/consvc-shepherd/blob/main/consvc_shepherd/settings.py#L131) and do:

```
logger = logging.getLoger(__name__)

logger.error("....")

File: openidc/middleware.py (hunk 2 above)
Comment (tiftran) [reply openidc/middleware.py]: was initially curious so i took a look at https://github.com/googleapis/google-auth-library-python/blob/main/google/oauth2/id_token.py to see what exception it raise. `verify_token()` calls `_fetch_certs()` and `decode()`. Glossed over `_fetch_certs()` and noticed `decode()` only raise ValueErrors. but nope... `_fetch_certs()` also raises an exception

File: openidc/middleware.py [hunk 5]
```diff
@@ -1,7 +1,33 @@
...
+    try:
+        decoded_jwt = id_token.verify_token(
+            iap_jwt,
+            requests.Request(),
+            audience=settings.IAP_AUDIENCE,
+            certs_url="https://www.gstatic.com/iap/verify/public_key",
+        )
+        return decoded_jwt["email"]
+    except Exception as e:
+        logger.error(f"**ERROR: JWT validation error {e}**'.")
```
ncloudioj, line 24:
```suggestion
        logger.error(f"IAP JWT validation error: {e}")
```

ncloudioj (approved):
r+ wc.

PR #81, opened 2023-01-16
chore: switch logs to stdout rather than stderr

tiftran:


consvc_shepherd/settings.py
```diff
@@ -129,6 +130,7 @@
             "level": env("SHEPHERD_ENV", default="DEBUG"),
             "class": "logging.StreamHandler",
             "formatter": "json",
+            "stream": sys.stdout,
```
tiftran, line 133: i think this does what it needs to do?

ncloudioj (approved):

PR #82, opened 2023-02-01
feat: add Shepherd Overview Page

tiftran:
Screen Shot 2023-01-30 at 8 57 26 PM

The overview page includes:

- a table view of the latest Snapshot.
- two select input so that a user can get the difference between to snapshots

The difference logic only accounts for the addition and or removal of advertisers,
I will put the advertiser hosts/paths logic as a follow up, as this is quite a bit already.

### DIFF
--- a/consvc_shepherd/forms.py
+++ b/consvc_shepherd/forms.py
@@ -1,0 +1,43 @@
+from django import forms
+
+from consvc_shepherd.models import SettingsSnapshot
+
+
+class SnapshotCompareForm(forms.Form):
+    older_snapshot = forms.CharField(
+        label="Older Snapshot",
+        widget=forms.Select(
+            choices=(
+                (snapshot.name, snapshot)
+                for snapshot in SettingsSnapshot.objects.all().order_by("-created_on")
+            )
+        ),
+    )
+    newer_snapshot = forms.CharField(
+        label="Newer Snapshot",
+        widget=forms.Select(
+            choices=(
+                (snapshot.name, snapshot)
+                for snapshot in SettingsSnapshot.objects.all().order_by("-created_on")
+            )
+        ),
+    )
+
+    def compare(self):
+        os_name = self.data["older_snapshot"]
+        ns_name = self.data["newer_snapshot"]
+        older_json_settings = SettingsSnapshot.objects.get(name=os_name).json_settings
+        newer_json_settings = SettingsSnapshot.objects.get(name=ns_name).json_settings
+
+        older_advertisers = set(older_json_settings["adm_advertisers"].keys())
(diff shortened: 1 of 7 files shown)


consvc_shepherd/forms.py
```diff
@@ -0,0 +1,43 @@
...
+        label="Older Snapshot",
+        widget=forms.Select(
+            choices=(
+                (snapshot.name, snapshot)
+                for snapshot in SettingsSnapshot.objects.all().order_by("-created_on")
+            )
+        ),
+    )
+    newer_snapshot = forms.CharField(
+        label="Newer Snapshot",
```
ncloudioj, line 17: Non-issue; for your consideration. How about we pre-select the most recent snapshot for "Newer Snapshot" and the 2nd most recent one for "Older Snapshot"? Also, if the user picks a snapshot as the "newer", the choices for "older" should be updated to only include those true older snapshots compared to "newer". Though that might require some JS ;)

consvc_shepherd/forms.py
```diff
@@ -0,0 +1,43 @@
...
+        os_name = self.data["older_snapshot"]
+        ns_name = self.data["newer_snapshot"]
+        older_json_settings = SettingsSnapshot.objects.get(name=os_name).json_settings
+        newer_json_settings = SettingsSnapshot.objects.get(name=ns_name).json_settings
+
+        older_advertisers = set(older_json_settings["adm_advertisers"].keys())
+        newer_advertisers = set(newer_json_settings["adm_advertisers"].keys())
+
+        added_advertisers = sorted(list(newer_advertisers - older_advertisers))
+        removed_advertisers = sorted(list(older_advertisers - newer_advertisers))
```
ncloudioj, line 36:
```suggestion
        added_advertisers = sorted(newer_advertisers - older_advertisers)
        removed_advertisers = sorted(older_advertisers - newer_advertisers)
```

templates/index.html
```diff
@@ -0,0 +1,69 @@
...
+                            {% endfor %}
+                        </table>
+                    </div>
+                {% endfor %}
+            {% endwith %}
+        </div>
+    </div>
+  </body>
+</html>
+
```
ncloudioj, line 67: Let's kill those trailing newlines.

ncloudioj (approved):
LGTM. r+ wc.

YEAH, the first non-admin page 🎉

consvc_shepherd/forms.py
tiftran: TIL that you can sort a set 😅

consvc_shepherd/forms.py
tiftran: oh that's fancy, probably ideal ux, but i think i'll leave this as another issue...

PR #89, opened 2023-03-20
[DISCO-2267] Added json schema for Shepherd Output

tiftran:
So... this doesn't achieve the ability to keep contile and shepherd in sync in terms of the json structure, but it's a start?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ COPY ./consvc_shepherd/ /app/consvc_shepherd/
 COPY ./static/ /app/static/
 COPY ./contile/ /app/contile/
 COPY ./openidc/ /app/openidc/
+COPY ./schema/ /app/schema/
 COPY ./manage.py /app/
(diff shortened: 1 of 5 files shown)


mtbottle (commented):
The code looks good, but I'm confused about how the schema is going to be used. Are we going to add a validate step before we upload the file? ie. before https://github.com/mozilla-services/consvc-shepherd/blob/037577957b131fdcb946c4e1ccc81558055da7a0/consvc_shepherd/admin.py#L24 Or is it meant to be checked manually?

mtbottle (approved):
> my original intent was that since there's a test related to the json schema, schema changes would be picked up before the code changes merged

Ah I see! I think this might be OK to have for now, as it protects us from changes to the application that affects the output format. Having it in the upload flow will ensure that the file will be correct format before we push the changes, which will be even better. Up to you if we want to add this!

mtbottle (approved):

PR #91, opened 2023-03-23
[DISCO-2304] fix: UI snapshot dropdowns update with new snapshots

tiftran:
Currently the compare snapshot ui dropdowns don't update when new setting snapshots are created, seems like setting choices like below doesn't dynamically get new snapshots
``` 
choices=(
                (snapshot.name, snapshot)
                for snapshot in SettingsSnapshot.objects.all().order_by("-created_on")
            ) 
```


ncloudioj (approved):
:shipit:

PR #93, opened 2023-04-14
[DISCO-2325] Create JSON Schema Validation for Control Mechanism

taddes:
## References

JIRA: [DISCO-2325](https://mozilla-hub.atlassian.net/browse/DISCO-2325)
GitHub: [#95  ](https://github.com/mozilla-services/consvc-shepherd/issues/95)

## Description
The goal of this task is to create the new JSON Schema for the Allocations of Kevel and adM partner advertisements. The document allows the distribution of ads by percentage supplying a percentage from 0 - 100 between the two and their occurrence in the array/list defines the Contile tiles from left to right, or the 0th index up.

The schema has description values for each key which supply sufficient context for anyone viewing the ticket.

Note: this task is only to create the valid schema. Subsequent tickets will cover the validation process, tests and potentially defining a model for this data.

A lengthy regex has been supplied to validate the title label in the format: `SOV-<UTC datetime:YYYYMMDDHHMMSS>`. For readability, here are the capture groups and their restrictions as it's not super readable in the JSON. We want an accurate representation of actual correct datetimes, hence the need to define multiple capture groups that ensure Y, M, D, h, m, s values are correct

```python
^(SOV-) 
([0-9]{4}) # year
(0[1-9]|1[0-2]) # month
(0[1-9]|[1-2][0-9]|3[0-1]) # day
(2[0-3]|[01][0-9]) # hour
([0-5][0-9]) # minute
([0-5][0-9])$ # second
```

See JSON Schema docs on regex matching [here](https://json-schema.org/understanding-json-schema/reference/string.html#pattern)

[DISCO-2325]: https://mozilla-hub.atlassian.net/browse/DISCO-2325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env.example
```diff
@@ -1,7 +1,10 @@
...
-OPENIDC_HEADER=
-IAP_AUDIENCE=
\ No newline at end of file
+DB_NAME=postgres
+DB_USER=postgres
+DB_HOST=localhost
+DB_PASS=postgres
+DJANGO_SETTINGS_MODULE=consvc_shepherd.settings
+OPENIDC_HEADER_PREFIX="accounts.google.com:"
+OPENIDC_HEADER="x-goog"
```
tiftran, line 10: lets move this out since it's included in your last PR

.gitignore
```diff
@@ -127,3 +127,10 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# DS_Store files for macOS
+**/.DS_Store
+
+# VS Code configuration files
+.vscode/*
+!.vscode/launch.json
```
tiftran, line 136: 👍

consvc_shepherd/admin.py
```diff
@@ -22,7 +22,7 @@ def publish_snapshot(modeladmin, request, queryset):
         snapshot.launched_by = request.user
         snapshot.launched_date = timezone.now()
         content = json.dumps(snapshot.json_settings, indent=2)
-        with open("./schema/shepherd.schema.json", "r") as f:
+        with open("./schema/filter.schema.json", "r") as f:
```
tiftran, line 25: actually can we rename it as adm_filter instead? sorry 😅

schema/test_schema.py
```diff
@@ -10,7 +10,7 @@ class JSONSchema(TestCase):
 
     def test_schema(self):
 
-        with open("./schema/shepherd.schema.json", "r") as f:
+        with open("./schema/filter.schema.json", "r") as f:
```
tiftran, line 13: same here please and thank you

schema/allocation.schema.json
```diff
@@ -0,0 +1,56 @@
+{
```
tiftran, line 1: Looks good! We'll add tests when Shepherd is able to produce json!

tiftran (approved):
Just some tiny things, but tyty for getting this done 🙂

consvc_shepherd/admin.py
taddes: Sure, wfm! 👍

.env.example
taddes: Sounds good! I may just match to what's on `main` now and there shouldn't be any conflicts when merging.

PR #94, opened 2023-04-19
[DISCO-2348] docs update

taddes:
Simple updates in `README` file for using pyenv virtualenv and some updates to running the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,17 @@ reducing the need for deployments of the services that use it.
 ## Quick Start
 
 To use consvc-shepherd, you'll need a Python 3.11 development environment and 
-poetry installed.
+poetry installed. To achieve this, it is recommended you use a virtual Python environment.
+
+venv is a viable option, but for ease you can use pyenv-virtualenv as a plugin for 
+pyenv and virtual environments. Using virtualenv creates a .python-version file  in the project directory and if you add `eval "$(pyenv virtualenv-init -)"` to your  `./bashrc` or `./zshrc`, the environment will automatically be activated.
+
+Ex:
+```shell
+export PATH="$HOME/.pyenv/bin:$PATH"
+eval "$(pyenv init -)"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,7 +7,21 @@ reducing the need for deployments of the services that use it.
...
+pyenv and virtual environments. Using virtualenv creates a .python-version file  in the project directory and if you add `eval "$(pyenv virtualenv-init -)"` to your  `./bashrc` or `./zshrc`, the environment will automatically be activated.
+
+Ex:
+```shell
+export PATH="$HOME/.pyenv/bin:$PATH"
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+```
+
+You also need to ensure you have Postgres (pgAdmin) installed. You can select a distribution from [https://www.postgresql.org/download/](https://www.postgresql.org/download/), use homebrew or their edb installer client.  Make sure to install version 14.7, or equivalent and have it running when developing for consvc-shepherd.
```
tiftran, line 22: I would exclude the mention of pgAdmin, considering there are other ways one could manager their db.

README.md
```diff
@@ -16,18 +30,36 @@ and rename it to `env`:
 ```shell
 DEBUG=true
 SECRET_KEY=keyboard-mash
+DB_NAME=postgres
+DB_USER=postgres
+DB_HOST=localhost
+DB_PASS=postgres
```
tiftran, line 36: Can we add these into .env.example as well?

tiftran (approved):
Thanks for the update! Just some considerations to make

README.md
taddes: Great point. I made a change that lists it as an option, but not the de-facto way, including cli or config files. Lmk if you think that's fine!

README.md
taddes: Yes! Good call

PR #101, opened 2023-04-21
[DISCO-2361] New Allocation File Capabilities

taddes:
Adding env variable in settings.py for a new file name, to support the new allocation json file while supporting previous adM filtering file.

Modify [storage.py](https://github.com/mozilla-services/consvc-shepherd/blob/main/consvc_shepherd/storage.py) to have an additional parameter to set a different file name.

Made change to base call of `send_to_storage` with parameter of `file_name` defined, so that current function is maintained and `GS_BUCKET_FILE_NAME` still supplied for adM filter file. 

A couple type annotations added as well. We discussed we'd like to add them as we go and make changes to have Shepherd fall more in line with mypy typing across the board.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,6 @@ dmypy.json
 # VS Code configuration files
 .vscode/*
 !.vscode/launch.json
+
+# DB Synchronization Stamps
+.install.stamp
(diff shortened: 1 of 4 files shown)


tiftran (approved):
Looks good tyty for the typing

PR #105, opened 2023-04-26
chore: add env so 'make test' works locally

tiftran:
@taddes sorry🙏

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,5 +31,5 @@ check: install
 
 .PHONY: test
 test: migration-check
-	export DJANGO_SETTINGS_MODULE=consvc_shepherd.settings && $(POETRY) run pytest --cov --cov-report=term-missing --cov-fail-under=$(COV_FAIL_UNDER)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,5 +31,5 @@ check: install
 
 .PHONY: test
 test: migration-check
-	$(POETRY) run pytest --cov --cov-report=term-missing --cov-fail-under=$(COV_FAIL_UNDER)
+	export DJANGO_SETTINGS_MODULE=consvc_shepherd.settings && $(POETRY) run pytest --cov --cov-report=term-missing --cov-fail-under=$(COV_FAIL_UNDER)
```
linabutler, line 34:
```suggestion
	env DJANGO_SETTINGS_MODULE=consvc_shepherd.settings $(POETRY) run pytest --cov --cov-report=term-missing --cov-fail-under=$(COV_FAIL_UNDER)
```

Make runs each line in its own sub-shell, anyway, so using `export` is okay, but `env` is a little shorter, and I think should also work! 😊

linabutler (approved):
Stamp! :shipit:

taddes (approved):
Thanks for fixing this @tiftran! LGTM!

PR #106, opened 2023-04-26
feat: added nav bar

tiftran:
Screenshot 2023-04-25 at 10 31 57 PM

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -2,12 +2,43 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8"/>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.10.4/font/bootstrap-icons.css">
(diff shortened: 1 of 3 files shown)


templates/base.html
```diff
@@ -2,12 +2,43 @@
...
+            <hr class="text-white">
+            <ul class="nav nav-pills flex-column mb-auto">
+                <li class="nav-item">
+                    <a href="/admin" class="nav-link text-white" aria-current="page">
+                        <i class="bi bi-person"></i>
+                        Admin
+                    </a>
+                </li>
+                <li class="nav-item">
+                    <a href="/" class="nav-link text-white" aria-current="page">
```
taddes, line 24: Just a question since I'm still getting familiar with Shepherd: The root is where the Snapshot Comparison view is presented, is that the same as the Adm filter, or is that filter something else within the Django admin page? Just curious if this was the href intended, or it'll be pointed at something down the road

taddes (approved):
LGTM @tiftran ! Great addition.

templates/base.html
tiftran: the index page is currently adm filter at the moment 🙂

PR #147, opened 2023-06-20
[DISCO-2434] feat: add allocation snaphot

tiftran:
## References

JIRA: [DISCO-2434](https://mozilla-hub.atlassian.net/browse/DISCO-2434)

## Description
We want to have the ability to view when allocations are published and by who, therefore introducing snapshots to allocations.

## PR Review Checklist

- [X] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/consvc-shepherd/blob/main/CONTRIBUTING.md).
- [X] The PR title starts with the JIRA issue reference, format `[DISCO-####]`, and has the same title (if applicable).

[DISCO-2434]: https://mozilla-hub.atlassian.net/browse/DISCO-2434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/admin.py
```diff
@@ -55,33 +56,36 @@ def publish_snapshot(modeladmin, request, queryset):
...
-            metrics.incr("allocation.upload.success")
-            messages.info(request, "Allocation setting has been published.")
-        except exceptions.ValidationError:
-            metrics.incr("allocation.schema.validation.fail")
-            messages.error(
-                request,
-                "JSON generated is different from the expected allocation schema. "
-                "Ensure that there are two allocation settings selected",
-            )
+    if len(queryset) > 1:
```
ncloudioj, line 59: Would `queryset` ever be empty? As we're tasting FP, here is the pattern match version for your consideration :)

```py
match queryset:
    case [query] | if query.launched_date is not None:
        messages.error(request, "... already been launched, ...")
    case [query]:
        # happy path
    case _:  # queryset is either empty or has more than 1 entry
       messages.error(request, "Only 1 snapshot can be published at the same time"
```

consvc_shepherd/admin.py
```diff
@@ -117,10 +120,64 @@ def has_delete_permission(self, request, obj=None) -> bool:
...
+        super(AllocationSettingsSnapshotModelAdmin, self).save_model(
+            request, obj, form, change
+        )
+        metrics.incr("allocations.snapshot.create")
+
+    def get_readonly_fields(self, request, obj=None) -> list:
+        """Return list of read-only fields for SettingsSnapshot."""
+        if obj:
+            return ["name"] + self.readonly_fields
+        return self.readonly_fields
```
ncloudioj, line 167:
```suggestion
        return ["name", *self.readonly_fields] if obj else self.readonly_fields
```

ncloudioj (approved):
Looks good, nice feature! 🚢

consvc_shepherd/admin.py
```diff
@@ -55,33 +56,36 @@ def publish_snapshot(modeladmin, request, queryset):
...
-                request,
-                "JSON generated is different from the expected allocation schema. "
-                "Ensure that there are two allocation settings selected",
-            )
+    if len(queryset) > 1:
+        messages.error(request, "Only 1 snapshot can be published at the same time")
+    elif queryset[0].launched_date is not None:
+        messages.error(
+            request,
+            "Snapshot has already been launched, create a new snapshot to launch",
```
taddes, line 64: I would suggest adding `Allocation` in front of snapshot so that the message is unique between the SettingsSnapshot and AllocationSettingsSnapshot panels.

consvc_shepherd/admin.py
```diff
@@ -117,10 +120,64 @@ def has_delete_permission(self, request, obj=None) -> bool:
 
     def delete_queryset(self, request, queryset) -> None:
         """Delete given SettingsSnapshot entry."""
-        super(ModelAdmin, self).delete_queryset(request, queryset)
+        super(SettingsSnapshotModelAdmin, self).delete_queryset(request, queryset)
         metrics.incr("filters.snapshot.delete")
 
 
+@admin.register(AllocationSettingsSnapshot)
+class AllocationSettingsSnapshotModelAdmin(admin.ModelAdmin):
```
taddes, line 128: As we discussed, the publication action to be on the side of the AllocationSettingSnapshot not on the AllocationSetting anymore, due to that attr error for `launched_date`

taddes (changes requested):
Great new feature @tiftran ! A couple things as we discussed regarding the publication action being on the side of publishing the snapshot, not the allocation, to prevent the `launched_date` attr issue. Thanks!

consvc_shepherd/admin.py
tiftran: yup, forgot to remove the action from the AllocationSettings side is all, good catch 🙂 tyty

consvc_shepherd/admin.py
tiftran: yessss flexing some FP in a real life scenario?

taddes (approved):
Great work @tiftran , just left one additional suggestion for your review about passing the ascending order to the JSON file in `save_model`.

PR #153, opened 2023-07-10
[DISCO-2441] allocation setting validators and boundaries

taddes:
## References

JIRA: [DISCO-2441](https://mozilla-hub.atlassian.net/browse/DISCO-2441)
GitHub: [#154 ](https://github.com/mozilla-services/consvc-shepherd/issues/154)

## Description
Adds some boundary checks to min and max values in admin console, does not by default show an extra input value when allocation already set, prevents the user from saving an allocation that is equal to 0, which is possible in admin and custom view at present.

Screen Shot 2023-07-10 at 1 34 09 PM

Screen Shot 2023-07-10 at 1 26 36 PM

Screen Shot 2023-07-10 at 1 25 53 PM

Just a note that we will want to open a ticket to add a delete option to the new custom allocation settings view. I discovered that even prior to this validation, you cannot delete an allocation in the custom UI.  If you select delete for the single extra field and the existing field, the validator kicks in and inhibits delete. See below:
Screen Shot 2023-07-10 at 1 59 00 PM

## PR Review Checklist

[DISCO-2441]: https://mozilla-hub.atlassian.net/browse/DISCO-2441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/admin.py
```diff
@@ -184,7 +184,7 @@ def delete_queryset(self, request, queryset) -> None:
 class PartnerAllocationInline(admin.TabularInline):
     """PartnerAllocationInline TabularInline child model."""
 
-    extra = 1
+    extra = 0
```
tiftran, line 187: I personally think leaving it is fine as is. Having a 1 there is just so that the partner form is already there when you create a new allocation. Since the allocation form is now moved out of the django admin, I foresee this admin form to be used only by engineers, so it's okay.

consvc_shepherd/forms.py
```diff
@@ -178,6 +178,15 @@ def clean(self):
...
 
+        if alive_forms:
+            if (
+                min((form.cleaned_data.get("percentage", 0) for form in alive_forms))
+                <= 0
+            ):
+                raise forms.ValidationError(
+                    "Allocation percentage must be greater than 0."
+                )
+
```
tiftran, line 189: I don't think we really need this changing [this](https://github.com/mozilla-services/consvc-shepherd/blob/main/consvc_shepherd/forms.py#L136) to 1 should be sufficient.
Screenshot 2023-07-10 at 9 41 22 AM

Future reference, I think it's best that we leave this `clean` method for validation related to the form as a whole and anything field specific in their own corresponding form specific `clean` methods

Makefile
```diff
@@ -23,7 +23,7 @@ black: $(INSTALL_STAMP)  ##  Run black
 
 .PHONY: flake8
 flake8: $(INSTALL_STAMP)  ##  Run flake8
-	$(POETRY) run flake8 $(APP_DIRS) --ignore=E203,E302,E501,E701
+	$(POETRY) run flake8 $(APP_DIRS) --ignore=E203,E302,E501,E701,W503
```
tiftran, line 26: we won't need this given that we don't need the form validation below

tiftran (changes requested):
We allowed 0% on the firefox side, I _think_ it should be okay if we leave out a partner, but I would need to double check about that, but given it is okay, there's some comments below relating to this.

consvc_shepherd/admin.py
taddes: Ok, wfm!

consvc_shepherd/tests/test_forms.py
```diff
@@ -152,3 +153,31 @@ def test_form_returns_true_when_valid(self):
...
+        position1_alloc: AllocationSetting = AllocationSetting.objects.create(
+            position=1
+        )
+        partner_allocation = PartnerAllocation.objects.create(
+            allocation_position=position1_alloc, partner=amp_partner, percentage=0
+        )
+        form = PartnerAllocationForm(partner_allocation.to_dict())
+        self.assertTrue(
+            partner_allocation.percentage < form.fields.get("percentage").min_value  # type: ignore [union-attr]
+        )
```
tiftran, line 183: Can we move these into a diff Test Class they shouldn't be in the same class with the SnapshotForm. In addition, I don't think this is effectively testing out the form. Could we structure them such that we have 
```
# .... data setup

form = PartnerAllocationForm(<data>)
self.assertFalse(form.is_valid())
# ... assertions with form.errors


```

tiftran (changes requested):

consvc_shepherd/tests/test_forms.py
taddes: Sure good call, thanks for the suggestion, I'll move it to its own class.

consvc_shepherd/tests/test_forms.py
```diff
@@ -152,3 +153,51 @@ def test_form_returns_true_when_valid(self):
...
+        self.partner_allocation_above = PartnerAllocation.objects.create(
+            allocation_position=self.position1_alloc,
+            partner=self.partner,
+            percentage=101,
+        )
+        self.partner_allocation_below = PartnerAllocation.objects.create(
+            allocation_position=self.position1_alloc, partner=self.partner, percentage=0
+        )
+        self.data_above: dict = {
```
tiftran, line 175: So data for a form would correspond to its fields, in this case, the `PartnerAllocationForm` data fields would be `partner` and `percentage`.

If debugged and step into this, after calling `form.is_valid()` i believe `form.errors` would return some error messages stating that fields are required.

tiftran (commented):
I've been flip flopping on this, and I think maybe it might be best to leave it as is for now . 
Since there are currently two partners at the moment and setting a partner to be 0 is harmless as firefox expects it, we also have validation to ensure 100% is allocated. We can revisit, if this becomes a problem when there are more partners.

I've left some fyis below

taddes:
Holding off on merging until discussing broader UI features and possible explicit partner definition.

PR #178, opened 2023-09-25
[DISCO-2574] Ability to run tests using docker workflow

tiftran:
## References

JIRA: [DISCO-2574](https://mozilla-hub.atlassian.net/browse/DISCO-2574)

## Description
Because we switch to a docker workflow, the existing `make test` does not work

## PR Review Checklist

[DISCO-2574]: https://mozilla-hub.atlassian.net/browse/DISCO-2574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,3 +73,7 @@ doc-preview: doc  ##  Preview Merino docs via the default browser
 .PHONY: dev
 dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
 	$(POETRY) run python manage.py runserver
+
+.PHONY: local_test
+local_test: $(INSTALL_STAMP)
+	docker-compose -f docker-compose.test.yml up --abort-on-container-exit
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,8 @@ doc-preview: doc  ##  Preview Merino docs via the default browser
 
 .PHONY: dev
 dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
-	$(POETRY) run python manage.py runserver
+	docker-compose up
 
-.PHONY: local_test
-local_test: $(INSTALL_STAMP)
+.PHONY: local-test
+local-test: $(INSTALL_STAMP)
 	docker-compose -f docker-compose.test.yml up --abort-on-container-exit
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -13,7 +13,7 @@ Unit tests are written and executed with `pytest` and are located in the test di
 
 ## Local Test Execution
 
-To execute unit tests in the local Docker container, run: `make local_test`.
+To execute unit tests in the local Docker container, run: `make local-test`.
 
 ## Note on CI Test Execution
 By default, unit tests run automatically in CI during each push to an active PR and during the merge process. 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -13,7 +13,7 @@ Unit tests are written and executed with `pytest` and are located in the test di
 
 ## Local Test Execution
 
-To execute unit tests locally, run: `make test`.
+To execute unit tests locally, run: `make local_test`.
 
 ## Note on CI Test Execution
 By default, unit tests run automatically in CI during each push to an active PR and during the merge process. 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -13,7 +13,7 @@ Unit tests are written and executed with `pytest` and are located in the test di
 
 ## Local Test Execution
 
-To execute unit tests locally, run: `make local_test`.
+To execute unit tests in the local Docker container, run: `make local_test`.
 
 ## Note on CI Test Execution
 By default, unit tests run automatically in CI during each push to an active PR and during the merge process. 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -72,4 +72,8 @@ doc-preview: doc  ##  Preview Merino docs via the default browser
 
 .PHONY: dev
 dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
-	$(POETRY) run python manage.py runserver
\ No newline at end of file
+	$(POETRY) run python manage.py runserver
+
+.PHONY: local_test
+local_test: $(INSTALL_STAMP)
+	docker-compose -f docker-compose.test.yml up --abort-on-container-exit
```
taddes, line 79: I think you'll want to remove the `--abort-on-container-exit` as it exits the entire application. You can reproduce by running shepherd in one terminal and then running the make command. 
```consvc-shepherd-db-1   | 2023-09-26 15:03:58.874 UTC [1] LOG:  received fast shutdown request
consvc-shepherd-db-1   | 2023-09-26 15:03:58.875 UTC [1] LOG:  aborting any active transactions
consvc-shepherd-db-1   | 2023-09-26 15:03:58.878 UTC [1] LOG:  background worker "logical replication launcher" (PID 32) exited with exit code 1
consvc-shepherd-db-1   | 2023-09-26 15:03:58.878 UTC [27] LOG:  shutting down
consvc-shepherd-db-1   | 2023-09-26 15:03:58.890 UTC [1] LOG:  database system is shut down
consvc-shepherd-db-1 exited with code 0```

However leaving it, it remains dangling and you can't access Shepherd. I'm looking into another solution, but thought I'd tag it in the review here in case you know an easy fix. Will follow up when I find something.  

We could alternatively state in docs that if someone is running the local Shepherd instance with Docker, that they should exit it or re-start after tests?

File: Makefile [hunk 2]
```diff
@@ -72,4 +72,8 @@ doc-preview: doc  ##  Preview Merino docs via the default browser
 
 .PHONY: dev
 dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
-	$(POETRY) run python manage.py runserver
\ No newline at end of file
+	$(POETRY) run python manage.py runserver
+
+.PHONY: local_test
```
taddes, line 77: Nice, I see you caught the `docker_test` rename 👍 Also, like below, perhaps `local-test` instead?

docs/testing.md
```diff
@@ -13,7 +13,7 @@ Unit tests are written and executed with `pytest` and are located in the test di
 
 ## Local Test Execution
 
-To execute unit tests locally, run: `make test`.
+To execute unit tests locally, run: `make local_test`.
```
taddes, line 16: This may be a little misleading, since it's not actually running tests locally but in the docker image. Perhaps rephrasing like: 
```suggestion
To execute unit tests in the local Docker container, run: `make local_test`.
```

My one other nit is that I see the Makefile convention is often to use a hyphen over an underscore (like in Merino), so perhaps `local-test` instead? Not really a big deal though.

taddes (changes requested):
Heyo @tiftran ! Thanks for adding this! I did add some comments as there is an issue of the Shepherd Docker instance stopping after the test runs. The others are just a few nits on naming. I'll follow up if I find something in the meantime, but thought to get this to you sooner.

Makefile
tiftran: This is intentional because yes, the db dangles. Though I don't see the need to access Shepherd after a test run. Is there a use case you're thinking of?

taddes: Yeah, for instance, if you're running shepherd, ideally running tests shouldn't make the application shut down. As is with Merino, we can run contract tests via docker against Merino without an active run of merino shutting down. We can indeed leave this be, I just think it would need to be documented so that a new person wouldn't be confused as to why their application crashed after running tests. The other thing is that sometimes ports will dangle and in this case, rebuilding and re-running shepherd will result in a port conflict which I encountered. You have to then kill the process, delete the containers and image, etc.

tiftran: So, I don't have problems needing to kill the process delete containers etc. Are you trying to `make dev` after? If that's the case,  I realized that wasn't changed when you were making changes to docs relating to moving into the docker workflow

taddes: No, not using `make dev`, `docker-compose build` then `docker-compose up`. But I see you made the update for dev to now use `docker-compose up`, so that does it!

docs/testing.md
```diff
@@ -13,7 +13,7 @@ Unit tests are written and executed with `pytest` and are located in the test di
 
 ## Local Test Execution
 
-To execute unit tests locally, run: `make test`.
+To execute unit tests in the local Docker container, run: `make local_test`.
```
taddes, line 16:
```suggestion
To execute unit tests in the local Docker container, run: `make local-test`.
```

taddes (approved):
Hey @tiftran . Thanks for updating `make dev` also. I wasn't using that to test or try since it was replaced by the docker workflow but makes sense to update it since we won't use the 'old fashioned way'. Sorry I never caught that  originally when documenting things.

approved, just one nit on renaming the docs from `local_test` to `local_test` 

🚀

PR #196, opened 2024-05-20
[AE-388] Add previews page

cbguder:
Adds a new page to show previews of Tiles and SPOCs.

Also fixes failing linters.

## References

JIRA: [AE-388](https://mozilla-hub.atlassian.net/browse/AE-388)

## Description

Screenshot 2024-05-20 at 2 44 29 PM

## PR Review Checklist

[AE-388]: https://mozilla-hub.atlassian.net/browse/AE-388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/preview.py
```diff
@@ -0,0 +1,196 @@
...
+        Region(code="NU", name="Nunavut"),
+        Region(code="YT", name="Yukon"),
+    ],
+}
+
+
+def get_spocs(country: str, region: str) -> list[Spoc]:
+    """Load SPOCs from MARS for given country and region"""
+    body = {
+        "pocket_id": "{03267ad1-0074-4aa6-8e0c-ec18e0906bfe}",
```
bbirdsong2, line 133: Should we generate a new pocket_id for each spocs request? That way we would avoid being frequency capped for ads.

cbguder: Good question! I frankly have no idea how this is used by Kevel. Do they use it to do frequency capping on the server side? If we generate a new one for each request, will that create a bunch of profiles on their side? Do they care? 😀

If you know more about this I'm happy to go with your suggestion.

dmueller: Frequency capping is the primary use case (for us) and https://dev.kevel.com/reference/request indicates if it isn't supplied then they'll generate one for the request, so it seems like they wouldn't care

consvc_shepherd/templatetags/sponsored_by.py
```diff
@@ -0,0 +1,26 @@
...
+
+LOCALIZATIONS = {
+    "US": "Sponsored by {sponsor}",
+    "CA": "Sponsored by {sponsor}",
+    "DE": "Werbung von {sponsor}",
+    "ES": "Patrocinado por {sponsor}",
+    "FR": "Sponsorisé par {sponsor}",
+    "GB": "Sponsored by {sponsor}",
+    "IT": "Sponsorizzata da {sponsor}",
+}
```
dmueller, line 15: curiosity - do you know if this is how Firefox does the localization too or a shortcut we're making here for the preview page since it's demoing the ads? I'd expect it's possible to use Firefox in en-US and be in Germany so you'd get ads for DE

dmueller (approved):

consvc_shepherd/templatetags/sponsored_by.py
cbguder: You are right, in Firefox you can have the interface language differ from your location, and it would use the interface language to localize this text. So you can indeed see "Sponsored by X" in English underneath ads for Germany.

I figured this is a reasonable shortcut to take for the preview page, instead of e.g. adding a separate dropdown for locale.

consvc_shepherd/preview.py
cbguder: Ah, alright fair enough, let me generate a new one here then.

While I'm at it, I think I'll also replace the Consumer Key with something bogus. MARS doesn't use it at all anyway.

PR #203, opened 2024-06-22
[AE-416] Show direct sold tiles on preview page

mashalifshin:
## References

JIRA: [AE-416](https://mozilla-hub.atlassian.net/browse/AE-416)

## Description

This PR adds adds "direct sold tiles" to the Tiles section on the Preview page, fetched by requesting `sponsored-topsite` placements from the `/spocs` endpoint. In this view, they are added to the right of the tiles coming from `/v1/tiles`.

Note that this new functionality is for showing the direct sold tiles when selecting Preview, Dev, and Prod environments from the dropdown. For unified API, those are pulled in on the backend already, and don't need to be put together by the client.

## PR Review Checklist

- [X] The PR title starts with the JIRA issue reference, format `[DISCO-####]`, and has the same title (if applicable).
- [X] [Documentation](https://github.com/mozilla-services/consvc-shepherd/tree/main/docs) has been updated (if applicable).

[AE-416]: https://mozilla-hub.atlassian.net/browse/AE-416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/consvc_shepherd/preview.py
+++ b/consvc_shepherd/preview.py
@@ -349,8 +349,7 @@ def get_ads(env: Environment, country: str, region: str) -> Ads:
         spocs_and_direct_sold_tiles = get_spocs_and_direct_sold_tiles(
             env, country, region
         )
-        print("spocs and direct sold tiles")
-        print(spocs_and_direct_sold_tiles)
+
         return Ads(
             tiles=amp_tiles + spocs_and_direct_sold_tiles[0],
             spocs=spocs_and_direct_sold_tiles[1],
--- a/consvc_shepherd/preview.py
+++ b/consvc_shepherd/preview.py
@@ -7,7 +7,6 @@ from typing import TypedDict
 import requests
 from django.views.generic import TemplateView
 
-
 # Localized strings from https://hg.mozilla.org/l10n-central/
 #
 # See e.g. https://hg.mozilla.org/l10n-central/es-ES/file/tip/browser/browser/newtab/newtab.ftl
@@ -258,7 +257,10 @@ def get_tiles(env: Environment, country: str, region: str) -> list[Tile]:
         for tile in r.json().get("tiles", [])
     ]
 
-def get_direct_sold_tiles(env: Environment, country: str, region: str) -> list[DirectSoldTile]:
+
+def get_direct_sold_tiles(
+    env: Environment, country: str, region: str
+) -> list[DirectSoldTile]:
     """Load Direct Sold Tiles (aka Sponsored Topsites) from MARS for given country and region"""
     # Generate a unique pocket ID per request to avoid frequency capping
     pocket_id = uuid.uuid4()
@@ -273,9 +275,9 @@ def get_direct_sold_tiles(env: Environment, country: str, region: str) -> list[D
             {
                 "name": "sponsored-topsites",
                 "zone_ids": [307565],
-                "ad_types": [2401, 3617]
+                "ad_types": [2401, 3617],
             }
-        ]
+        ],
     }
 
     r = requests.post(f"{env.mars_url}/spocs", json=body, timeout=30)
@@ -332,11 +334,7 @@ def get_unified(env: Environment, country: str) -> Ads:
         for spoc in r.json().get(spocs_placement, [])
     ]
 
-    return Ads(
-        spocs=spocs,
-        tiles=tiles,
-        direct_sold_tiles=[]
-    )
+    return Ads(spocs=spocs, tiles=tiles, direct_sold_tiles=[])
 
 
 def get_ads(env: Environment, country: str, region: str) -> Ads:
@@ -347,7 +345,7 @@ def get_ads(env: Environment, country: str, region: str) -> Ads:
         return Ads(
             spocs=get_spocs(env, country, region),
             tiles=get_tiles(env, country, region),
-            direct_sold_tiles=get_direct_sold_tiles(env, country, region)
+            direct_sold_tiles=get_direct_sold_tiles(env, country, region),
         )
 
 
--- a/consvc_shepherd/preview.py
+++ b/consvc_shepherd/preview.py
@@ -3,7 +3,7 @@
 import uuid
 from dataclasses import dataclass
 from typing import TypedDict
-from urllib.parse import parse_qs, urlparse, urlencode, urlunparse
+from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 import requests
 from django.views.generic import TemplateView
@@ -33,6 +33,7 @@ LOCALIZATIONS = {
     },
 }
 
+DIRECT_SOLD_TILE_AD_TYPE = 3120
 
 class Region(TypedDict):
     """Represents a supported country or region within a country
@@ -52,6 +53,7 @@ class Environment:
     name: str
     mars_url: str
     spoc_site_id: int | None
+    direct_sold_tile_zone_id: int | None
 
 
 @dataclass(frozen=True)
@@ -89,30 +91,35 @@ ENVIRONMENTS: list[Environment] = [
         name="Dev",
         mars_url="https://mars.stage.ads.nonprod.webservices.mozgcp.net",
         spoc_site_id=1276332,
+        direct_sold_tile_zone_id=None
     ),
     Environment(
         code="preview",
         name="Preview",
         mars_url="https://mars.prod.ads.prod.webservices.mozgcp.net",
         spoc_site_id=1084367,
+        direct_sold_tile_zone_id=None,
     ),
     Environment(
         code="production",
         name="Production",
         mars_url="https://mars.prod.ads.prod.webservices.mozgcp.net",
         spoc_site_id=1070098,
+        direct_sold_tile_zone_id=280143
     ),
     Environment(
         code="unified_dev",
         name="Dev Unified API",
         mars_url="https://mars.stage.ads.nonprod.webservices.mozgcp.net",
         spoc_site_id=None,
+        direct_sold_tile_zone_id=None,
     ),
     Environment(
         code="unified_prod",
         name="Prod Unified API",
         mars_url="https://mars.prod.ads.prod.webservices.mozgcp.net",
         spoc_site_id=None,
+        direct_sold_tile_zone_id=None,
     ),
 ]
 
@@ -213,6 +220,8 @@ def get_spocs(env: Environment, country: str, region: str) -> list[Spoc]:
     }
 
     r = requests.post(f"{env.mars_url}/spocs", json=body, timeout=30)
+    print('spocs from /spocs')
+    print(r.json())
 
     return [
         Spoc(
@@ -241,6 +250,9 @@ def get_tiles(env: Environment, country: str, region: str) -> list[Tile]:
         f"{env.mars_url}/v1/tiles", params=params, headers=headers, timeout=30
     )
 
+    print('tiles from /v1/tiles')
+    print(r.json())
+
     return [
         Tile(
             image_url=tile["image_url"],
@@ -254,9 +266,11 @@ def get_tiles(env: Environment, country: str, region: str) -> list[Tile]:
 def get_direct_sold_tiles(
     env: Environment, country: str, region: str
 ) -> list[Tile]:
-    """Load Direct Sold Tiles (aka Sponsored Topsites) from MARS for given country and region"""
+    """Request Direct Sold Tiles (aka Sponsored Topsites) from MARS for given country and region"""
     # Generate a unique pocket ID per request to avoid frequency capping
     pocket_id = uuid.uuid4()
+
+    direct_sold_tiles_placement="sponsored-topsites"
 
     body = {
         "pocket_id": f"{{{pocket_id}}}",  # produces "{uuid}"
@@ -266,14 +280,17 @@ def get_direct_sold_tiles(
         "region": region,
         "placements": [
             {
-                "name": "sponsored-topsites",
-                "zone_ids": [307565],
-                "ad_types": [2401, 3617],
+                "name": direct_sold_tiles_placement,
+                "zone_ids": [env.direct_sold_tile_zone_id],
+                "ad_types": [DIRECT_SOLD_TILE_AD_TYPE],
             }
         ],
     }
 
     r = requests.post(f"{env.mars_url}/spocs", json=body, timeout=30)
+
+    print('direct sold tiles aka sponsored-topsites from /spocs')
+    print(r.json())
 
     return [
         Tile(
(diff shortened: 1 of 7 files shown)


consvc_shepherd/preview.py
```diff
@@ -64,6 +64,13 @@ class Spoc:
     sponsored_by: str
 
 
+@dataclass(frozen=True)
+class DirectSoldTile(Spoc):
```
mashalifshin, line 68: I am using the same model as `Spoc` since that data is so similar, but lmk if that should change

consvc_shepherd/preview.py
```diff
@@ -250,6 +258,42 @@ def get_tiles(env: Environment, country: str, region: str) -> list[Tile]:
...
+    body = {
+        "pocket_id": f"{{{pocket_id}}}",  # produces "{uuid}"
+        "site": env.spoc_site_id,
+        "version": 2,
+        "country": country,
+        "region": region,
+        "placements": [
+            {
+                "name": "sponsored-topsites",
+                "zone_ids": [307565],
```
mashalifshin, line 277: Currently using defaults I found in the [MARS backend kevel config](https://github.com/mozilla-services/mars/blob/5f257b7e3110baeb2764598f3ab6c183b92e21fb/config/kevel.go) just to get started, but we'll need real values here for `zone_ids` and `ad_types`

consvc_shepherd/preview.py
```diff
@@ -290,10 +334,7 @@ def get_unified(env: Environment, country: str) -> Ads:
         for spoc in r.json().get(spocs_placement, [])
     ]
 
-    return Ads(
-        spocs=spocs,
-        tiles=tiles,
-    )
+    return Ads(spocs=spocs, tiles=tiles, direct_sold_tiles=[])
```
mashalifshin, line 337: Lmk if I am missing something, but it [seems like unified API doesn't yet support a placement for direct sold tiles](https://github.com/mozilla-services/mars/blob/5f257b7e3110baeb2764598f3ab6c183b92e21fb/unified_ads_api/handler.go#L14), so for now it will always return an empty list.

consvc_shepherd/preview.py
```diff
@@ -250,6 +258,42 @@ def get_tiles(env: Environment, country: str, region: str) -> list[Tile]:
     ]
 
 
+def get_direct_sold_tiles(
```
mashalifshin, line 261: This call is strikingly similar to the existing call to `/spocs`, and since we are hitting the same endpoint, it should be possible to get all the data in just one call. 

To do that, we'd need to figure out what the backend's current defaults are for the `placements` paremeter for the spocs call, so that we could DRY up the calls. Not sure if that's necessary here, since it is just an internal preview tool, an extra call doesn't seem high impact.

docker-compose.yml
```diff
@@ -1,7 +1,6 @@
-version: "3"
-
 services:
   db:
+    container_name: consvc-shepherd-db
```
mashalifshin, line 3: This small tweak saves time when exec-ing into the container -- you can do it by name instead of querying for PID. I think this `docker-compose` file is only used for development, so it's okay to make this change. But if this is used in other environments that rely on its automated container name, this change could cause trouble

docs/quickStart.md
```diff
@@ -16,12 +16,12 @@ eval "$(pyenv virtualenv-init -)"
 ```
 
 4. To run consvc-shepherd, you'll need to specify some minimal configurations.
-Use the existing `.env.example` and rename it to `env`.
+Use the existing `.env.example` and copy it to `.env`.
```
mashalifshin, line 19: This tiny changes keeps the git tree clean when managing the `.env` file

docs/quickStart.md
```diff
@@ -40,33 +40,29 @@ $ pyenv install 3.11
...
-# Install dependencies
-$ pip install poetry
-$ poetry install
+$ pyenv local shepherd # enables virtual env when you enter directory.
 ```
 
 6. Install your dependencies:
+
 ```shell
+$ pip install poetry
```
mashalifshin, line 49: Tiny reorganizing for clarity

templates/preview.html
```diff
@@ -73,6 +73,23 @@ <h2>SPOCs</h2>
             </div>
         {% endfor %}
     </div>
+
+    <h2>Direct Sold Tiles</h2>
+    <div class="spocs">
```
mashalifshin, line 78: Reusing the `spocs` classes here, as it seems the direct sold tiles look the same, but please lmk if I'm mistaken.

docker-compose.test.yml
```diff
@@ -1,5 +1,3 @@
-version: "3"
```
mashalifshin, line 1: No longer need version here, this change cleans up a warning

consvc_shepherd/tests/test_views.py
```diff
@@ -142,3 +145,49 @@ def test_view_returns_post_request(self):
         )
         self.assertEqual(response.status_code, 200)
         self.assertEqual(AllocationSettingsSnapshot.objects.count(), 1)
+
+
+@override_settings(DEBUG=True)
+class TestPreviewView(TestCase):
```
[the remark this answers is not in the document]
mashalifshin: Done.

docker-compose.yml
mashalifshin: Didn't get a chance to track this down, but I removed this change from the PR while I figure this out.

consvc_shepherd/preview.py
```diff
@@ -326,6 +368,18 @@ def find_env_by_code(env_code: str) -> Environment:
     raise ValueError(f"Unknown environment '{env_code}'")
 
 
+def resize_direct_sold_tile_image_url(img_url: str, w: int, h: int) -> str:
```
cbguder, line 371: It might come out to the same thing in practice, but I would prefer to replicate Firefox's behavior exactly here, for 2 reasons:
1. This would break if we start returning MARS Image Proxy URLs in `image_src` (instead of Pocket Image Proxy URLs like we do today)
2. It might also diverge from FF behavior if we add/remove query params to `image_src`

(Admittedly both of those are unlikely)

FF uses `raw_image_src` and constructs the URL like this: https://hg.mozilla.org/mozilla-central/file/tip/browser/components/newtab/lib/TopSitesFeed.sys.mjs#l1058

consvc_shepherd/preview.py
```diff
@@ -88,30 +92,35 @@ class Ads:
         name="Dev",
         mars_url="https://mars.stage.ads.nonprod.webservices.mozgcp.net",
         spoc_site_id=1276332,
+        direct_sold_tile_zone_id=317832,  # In Kevel > Network: MozAds-Dev, Site: Firefox Experiments corollary, Zone: Tiles"
```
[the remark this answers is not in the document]
dmueller: I tried to create all the same site/zone setup under the dev network that we actually use in production, so my "Tiles" zone mentioned here is within the dev network's production corollary site. I should have called out the id I was referring to to avoid that ambiguity -- 317828

consvc_shepherd/preview.py
```diff
@@ -250,6 +259,40 @@ def get_tiles(env: Environment, country: str, region: str) -> list[Tile]:
     ]
 
 
+def get_direct_sold_tiles(env: Environment, country: str, region: str) -> list[Tile]:
```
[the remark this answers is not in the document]
mashalifshin: > Nit: This function could just return early if direct_sold_tile_zone_id is None.

Good call, done! 

> It could also be merged with get_spocs (as the /spocs endpoint is able to return multiple placements) but that is an optional optimization and no big deal IMO if we keep them separate.

Yes indeed, this would be better and faster. I decided to punt on it because to achieve it I'd need to track down the current default placements (the placements are not explicit in`get_spocs`: `# Omit placements to use server-side defaults`). And 'cause if I'm understanding correctly, we'd like to remove this way of requesting tiles soon in favor of the unified API calls anyway.

Let me know what you think is best
- leave as is
- track down the `get_spocs` placements values and add to this PR
- track down the `get_spocs` placements values to add to a separate, fast-follow PR
- other ideas?

consvc_shepherd/preview.py
```diff
@@ -290,20 +333,19 @@ def get_unified(env: Environment, country: str) -> Ads:
...
-    )
+    return Ads(spocs=spocs, tiles=tiles)
 
 
 def get_ads(env: Environment, country: str, region: str) -> Ads:
     """Based on Environment, either load spocs & tiles individually or from a single request"""
     if env.code.startswith("unified_"):
         return get_unified(env, country)
     else:
+        tiles = get_tiles(env, country, region)
```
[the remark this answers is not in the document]
mashalifshin: I dig it! I was thinking the same thing, but wasn't sure what a good name would be. 

(Sidebar: I think we gotta refine/create our own terminology as we shape our own product here. Some things seem to be named because they come from a particular 3rd party (amp tiles), others from a sales strategy (direct sold tiles). Perhaps in our own system these are called "newtab tiles"? (not sure if a good name, just an example). Unified API feels like an opportunity for this... )

mashalifshin (commented):
Thanks for the review @cbguder and @dmueller. I've addressed the quick feedback, and I've got a little more work to do with constructing image URLs, so I'll let you know when it's ready for another look.

consvc_shepherd/preview.py
mashalifshin: Ooh, thank you for reiterating this! I mistakenly thought that since I was no longer trying to fit spoc-shaped images into tile-shaped containers on the page, that this was no longer necessary. But yeah I totally see that it's important to render here the same way as FF, I'll push up a new version soon

consvc_shepherd/preview.py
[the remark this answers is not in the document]
cbguder: Yeah, that's my bad for being lazy and defaulting to the server-side settings =)

The values are not hard to find (they're in MARS code & config) but I think it's OK to leave this as is for now.

consvc_shepherd/preview.py
```diff
@@ -88,30 +92,35 @@ class Ads:
...
         mars_url="https://mars.stage.ads.nonprod.webservices.mozgcp.net",
         spoc_site_id=1276332,
+        direct_sold_tile_zone_id=317832,  # In Kevel > Network: MozAds-Dev, Site: Firefox Experiments corollary, Zone: Tiles"
     ),
     Environment(
         code="preview",
         name="Preview",
         mars_url="https://mars.prod.ads.prod.webservices.mozgcp.net",
         spoc_site_id=1084367,
+        direct_sold_tile_zone_id=317828,  # In Kevel > Newtork: MozAds-Dev, Site: Firefox Production corollary, Zone: Tiles
```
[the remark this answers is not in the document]
dmueller: I also only have read-only there, we'll need to ask Laura to make it for us

consvc_shepherd/preview.py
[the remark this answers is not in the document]
cbguder: Yeah, we had a conversation around this a few months ago: https://mozilla.slack.com/archives/C05HFHMRKCN/p1712258948038509

Unified API uses the term "provider" for these and Kevel and AMP are providers, i.e. the entities serving the ads.

Separately, there is the entity that sold the inventory. In the case of Kevel, that is Mozilla (aka "Direct Sold" or "moz-sales"). For AMP, that is still AMP, so they perform both roles.

consvc_shepherd/preview.py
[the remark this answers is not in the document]
cbguder: You can double check with Laura, but I think they just don't use zones in preview, so what ads we get back are really determined by site ID & ad type.

For instance I can make a request to Kevel for network 10250, site 1084367, adType 3120 and get direct sold tiles, regardless of zone ID

consvc_shepherd/preview.py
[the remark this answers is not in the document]
mashalifshin: Thanks for tracking down those values for me, now the PR does use a single call for both placements 🙌

consvc_shepherd/preview.py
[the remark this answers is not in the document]
mashalifshin: Nice, thank you for these

consvc_shepherd/preview.py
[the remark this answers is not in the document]
mashalifshin: I'm planning to ask Laura for a tour of Kevel sometime next week and I will check with her then. Just generally need to understand how we use the envs and Kevel structures better...

consvc_shepherd/preview.py
mashalifshin: Okay, I think I've now replicated the FF behavior here

mashalifshin:
@cbguder @dmueller Whenever you've got a moment, this PR is ready for another look.

consvc_shepherd/tests/test_preview.py
```diff
@@ -0,0 +1,105 @@
...
+                self.assertEqual(ads.spocs[0].sponsored_by, SPOC.sponsored_by)
+
+                # Tile data
+                self.assertEqual(
+                    ads.tiles[2].image_url, "https://picsum.photos/id/40/48/48"
+                )
+                self.assertEqual(ads.tiles[2].name, "Progress Quest")
+                self.assertEqual(ads.tiles[2].sponsored, "Sponsored")
+
+    def test_get_ads_unified_env(self):
```
cbguder, line 92: Nit: this test doesn't test much in terms of behavior. I think we can extend it or drop it.

consvc_shepherd/tests/test_preview.py
```diff
@@ -0,0 +1,105 @@
+"""Unit tests for the preview page functionalities"""
+
+from unittest import mock
+
+from django.test import TestCase, override_settings
+
+from consvc_shepherd.preview import Environment, Spoc, Tile, get_ads
+
+SPOC: Spoc = Spoc(
```
cbguder, line 9: Nit: I think the type annotations for these constants are not necessary

cbguder (approved):
LGTM! Thank you!

consvc_shepherd/tests/test_preview.py
mashalifshin: Yeah I see what you're saying. I think what I really want here is a "view test" sort of thing that verifies the controls on the page are working and doing the right thing. So this is a small gesture towards that, which verifies we're calling the right things for the unified api calls. Though maybe not verifiying it very well, perhaps also asserting the the un-unified api methods are not called would help?

But yeah generally maybe efforts are better put towards more unit tests of the helper functions in `preview.py`, or doing a proper view test. I'd welcome any additional thoughts you have, I'm going to merge this as-is but I'll be continuing to revisit the test strategy as I keep building here ...

consvc_shepherd/tests/test_preview.py
mashalifshin: Fixed. Good call, thank you!

PR #205, opened 2024-07-08
include user-agent header when requesting ads from unified api

dmueller:
## Description
ADM tiles aren't showing up on the preview page. This is a side effect of the ADM provider returning no tiles based on the user agent of the request.

## Solution
Include a user-agent header to specify Firefox. Ran locally and confirmed dev unified api & prod unified api now include tiles from ADM. dev shows the test kevel tile campaign, but prod does not reliably have a kevel tile -- that is much more dependent on inventory/country of the request.

Screenshot 2024-07-08 at 11 31 10 AM

## PR Review Checklist

- [X] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/consvc-shepherd/blob/main/CONTRIBUTING.md).
- [X] [Documentation](https://github.com/mozilla-services/consvc-shepherd/tree/main/docs) has been updated (if applicable).
- [X] Functional and performance test coverage has been expanded and maintained (if applicable).

### DIFF
--- a/consvc_shepherd/preview.py
+++ b/consvc_shepherd/preview.py
@@ -259,6 +259,11 @@ def get_unified(env: Environment, country: str) -> Ads:
     spocs_placement = "newtab_spocs"
     tiles_placement = "newtab_tiles"
 
+    # specify Firefox user-agent header so that newtab_tiles are filled
+    headers = {
+        'user-agent':'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:129.0) Gecko/20100101 Firefox/129.0'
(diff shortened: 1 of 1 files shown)


cbguder (approved):
Not opposed to landing this change to get things working, but we need to drop this requirement for Unified API before we go live, and I think we can safely drop it for /v1/tiles as well.

dmueller:
We chatted on slack & the user-agent expectation in MARs isn't adding much value since bad actors could also include a user-agent header similar to this adjustment; so going to go the route of removing that check instead.

PR #209, opened 2024-07-16
AE-424, 425, 426, and debugging report feature

tanfarming:
## References

JIRA: 
- https://mozilla-hub.atlassian.net/browse/AE-424
- https://mozilla-hub.atlassian.net/browse/AE-425
- https://mozilla-hub.atlassian.net/browse/AE-426

## Description
- added regions for all countries
- made ads clickable
- added mobile support for /tiles and /spocs
- added a preliminary debugging text box, atm just reporting parameters and result

## PR Review Checklist


consvc_shepherd/tests/test_preview.py
```diff
@@ -12,25 +12,30 @@
     domain="cosmetics.beauty",
     excerpt="The sale has begun...",
     sponsored_by="Sponsored by Cosmetics",
+    url="example.com",
```
dmueller, line 15: given we have assert.eq checking the url value, it would be good to have different values for different test data

consvc_shepherd/preview.py
```diff
@@ -340,14 +332,14 @@ def get_unified(env: Environment, country: str) -> Ads:
...
+def get_ads(env: Environment, country: str, region: str, agent: str) -> Ads:
     """Based on Environment, either load spocs & tiles individually or from a single request"""
     if env.code.startswith("unified_"):
         return get_unified(env, country)
     else:
-        amp_tiles = get_amp_tiles(env, country, region)
+        amp_tiles = get_amp_tiles(env, country, region, agent)
         spocs_and_direct_sold_tiles = get_spocs_and_direct_sold_tiles(
-            env, country, region
+            env, country, region, "Mobile" in agent
```
dmueller, line 342: `"Mobile" in agent` looks like it is consistent with the user-agent parsing in mars, but at first glance I wondered if this wouldn't line up since the parsing doesn't occur in that way (it uses woothee parser and then looks at some of the fields that extracts). We could similarly avoid any concern here using a representative mobile user-agent.

consvc_shepherd/preview.py
```diff
@@ -132,11 +143,28 @@ class Ads:
...
+    "Mozilla/5.0 (Android 11; Mobile; rv:92.0) Gecko/92.0 Firefox/92.0",
+    "iPad; CPU iPhone OS 11_5_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/35.0 Mobile/15E148 Safari/605.1.15",
+    "Mozilla/5.0 (iPhone; CPU iPhone OS 11_5_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/35.0 Mobile/15E148 Safari/605.1.15",
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:10.0) Gecko/20100101 Firefox/91.0",
+    "Mozilla/5.0 (iPhone; CPU iPhone OS 14_8_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/91.0 Mobile/15E148 Safari/605.1.15",
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36",
+    "GoogleStackdriverMonitoring-UptimeChecks(https://cloud.google.com/monitoring)",
+    "python-requests/2.26.0",
+    "curl/8.7.1",
+]
```
dmueller, line 166: Thoughts on having the dropdown be a selection between mobile / desktop rather than a user-agent selector and then the preview page using a representative user-agent value for the corresponding selection?

I wouldn't expect the ad-ops team to know which of the user agent strings would correlate to mobile vs desktop, so if they are the customers of the page the full user-agent strings might not be as helpful.

templates/preview.html
```diff
@@ -81,6 +105,8 @@ <h2>SPOCs</h2>
     const countrySelect = document.getElementById("country");
     const regionContainer = document.getElementById("region-container");
     const regionSelect = document.getElementById("region");
+    const agentSelect=document.getElementById('agent')
+    const mobileLabel=document.getElementById('mobileLabel')
```
dmueller, line 109: does the `mobileLabel` Id exist? doing a search looks like this is the only reference

consvc_shepherd/preview.py
tanfarming: yeah i was thinking about getting a freebie of a potential "custom user agent" feature in future, can come in handy for testing.
the list of user agents were copied from [mars' device_test](https://github.com/mozilla-services/mars/blob/main/tiles/device_test.go), including the invalid ones
on `"Mobile" in agent` -- yeah in mars it's determined by checking woothee parsed formFactor and osFamily, but here since we are not really dealing with real/organic/potentially arbitrary user-agents, i thought to keep things simple and just used the ones that used for test in mars, the `Mobile` in agent is really there to make it easier to read, as well as avoid the complexity of adding and passing around an extra isMobile variable

tldr, we are just facing user-agents here so i thought I'd favor simplicity, but yeah if you feel that the check should be made more "solid" or "real", i think i can do one of the 2 things
- put the check into a function to have more control in future
- hardcode the isMobile bits into the agents struct to avoid the need to make a decision
- make it real, mirroring the check in shepherd, aka woothee parse and do the formFactor/osFamily checks, although i don't see a good reason for paying this bit of extra complexity?

let me know what you think and i'll get the code updated

consvc_shepherd/preview.py
tanfarming: ah, thanks for pointing this out, i'll make some changes to make it easier to tell

templates/preview.html
tanfarming: oh that's an idea i tried that didn't make it into the final version, i'll drop it, thanks for point it out!

consvc_shepherd/preview.py
cbguder: +1 on @dmueller's comment. Let's not forget we (Ads Eng) are not the primary users of this page. The primary user is Ad Ops.

templates/preview.html
```diff
@@ -40,26 +52,37 @@ <h1>MARS Ads Preview</h1>
         <br/><br/>
 
         <input type="submit" value="Reload"/>
+        <br/><br/>
+        <div class="container shadow p-3 mb-3" style="margin-top:3%">
```
cbguder, line 56: In line with my comment above, could we maybe have this at the bottom of the page?

tanfarming: yep done

consvc_shepherd/preview.py
tanfarming: cc @cbguder ^

cbguder: I think it should literally just be Mobile and Desktop. Two choices.

The goal of this page is not to debug MARS. It is to enable Ad Ops to preview ads.

templates/preview.html
cbguder: WDYT about also hiding it by default and having a button/checkbox/etc. to show it or something?

consvc_shepherd/preview.py
tanfarming: hmm how about just 2 groups for mobile and desktop, with android and iphone in mobile, and mac/windows/linux in desktop? 
image

consvc_shepherd/tests/test_preview.py
tanfarming: done

consvc_shepherd/preview.py
tanfarming: reasoning behind's that, with this config i think the risk of confusion should be minimal? and we are still covered if in future we further branching into different OS

templates/preview.html
tanfarming: done

consvc_shepherd/preview.py
dmueller: It's more a simplicity thing than worrying about future proofing -- currently we sell ad inventory in one bucket on desktop, and another bucket on mobile. The fact that is derived from a user agent is specific to this case, but what we're really trying to do is give visibility into the various ad inventory buckets.

So from that perspective "Mobile / Desktop" is the distinction from a sales and ad operations management perspective.

tanfarming: yep, np, done
latest:
image

templates/preview.html
```diff
@@ -21,6 +21,27 @@ <h1>MARS Ads Preview</h1>
...
+                        <option value="{{ anAgent }}">Mobile</option>
+                    {% endif %}
+                {% elif 'Firefox' in anAgent and 'Mobile' not in anAgent %}
+                    {% if anAgent == agent %}
+                        <option value="{{ anAgent }}" selected="selected">Desktop</option>
+                    {% else %}
+                        <option value="{{ anAgent }}">Desktop</option>
+                    {% endif %}
+                {% endif %}
+            {% endfor %}
```
dmueller, line 41: this could be simplified if it followed the pattern of other dropdowns, where the list of options contained a field for each of the value and the displayed text, e.g. ENVIRONMENTS or COUNTRIES has .code/.name.

```AGENTS: list[Agent] = [
  Agent(
    code="Mozilla/5.0 (Windows NT 10.0; rv:10.0) Gecko/20100101 Firefox/91.0",
    name="Desktop"
  ),
  Agent(
    code="Mozilla/5.0 (Android 11; Mobile; rv:92.0) Gecko/92.0 Firefox/92.0",
    name="Mobile"
  )
]```

in typing this out I think it would also make sense to have a boolean field to indicate if the agent is_mobile on the Agent class; with that then nothing would need to test with `Mobile in...` since we have a static set of them and already know what they are.

this would be similar to passing in the extra fields how ENVIRONMENT works -- given a code, it looks up which Environment that is and then passes that Environment in to the ad fetch method

File: consvc_shepherd/preview.py (hunk 3 above)
Comment (dmueller) [reply consvc_shepherd/preview.py]: awesome, looking good! i left a couple smaller comments just then but i like where this is ending up

File: templates/preview.html (hunk 6 above)
Comment (tanfarming) [reply templates/preview.html]: done https://github.com/mozilla-services/consvc-shepherd/compare/3c751a03189b60f96edc4c5c8cf391f78b817243..02d2424fe526229e9f02cb458680b95f1bef0b2e

File: consvc_shepherd/preview.py [hunk 7]
```diff
@@ -403,17 +384,29 @@ def get(self, request, *args, **kwargs):
...
 
         env = find_env_by_code(env_code)
 
+        debugMsg = f"{' &#x1F4F1; ' if 'Mobile' in agent else ''}country: {country}, region: {region}<br>env: {env}<br>{agent}"
+        try:
+            ads = get_ads(env, country, region, agent)
+            debugMsg += "<br>&#x2705;ok"
+        except Exception as e:
+            ads = {}
+            debugMsg += f"<br>&#x1F480;&#x1F480;&#x1F480; (bad User Agent?) <br> {e}, <br> {traceback.format_exc()}"
```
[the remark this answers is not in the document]
tanfarming: done

dmueller (approved):

PR #215, opened 2024-07-29
[AE-458] Import Boostr Deals and Products into Shepherd

mashalifshin:
## References

JIRA: [AE-458](https://mozilla-hub.atlassian.net/browse/AE-458)

## Description

This change adds a new manual script, `sync_boostr_data`, which can call the Boostr API and upserts the latest copy of Deals and Products to shepherd. The script also inserts the monthly budgets for each product in the deal, so that we can start building the Inventory Overview and Revenue Overview.

Note that this script does not yet pull in CPC/CPM rates for the deals, which we'll need to add eventually in order to calculate values in the Campaign Overview. Captured in https://mozilla-hub.atlassian.net/browse/AE-504

Still to do on this branch:
- Unit tests for sync script
- See if I can move the script out of the root (currently getting an error setting up the Django environment settings if I try this)

## PR Review Checklist

[AE-458]: https://mozilla-hub.atlassian.net/browse/AE-458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Makefile
```diff
@@ -64,16 +64,16 @@ doc-install-deps:  ## Install the dependencies for doc generation
...
+	mdbook-mermaid install && mdbook clean && mdbook build
 
 .PHONY: doc-preview
 doc-preview: doc  ##  Preview Merino docs via the default browser
 	mdbook serve --open
 
 .PHONY: dev
 dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
-	docker-compose up
+	docker compose up
```
mashalifshin, line 75: The command is now `docker compose` as of the latest version, so updated this throughout

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
+from dataclasses import dataclass
+import django
+import environ
+import logging
+import math
+import os
+import requests
+
```
cbguder, line 8: Nit: the Python convention is to have 2 blank lines between top-level definitions.

Of all the linting we have on this repository, I'm pretty surprised [we chose to ignore](https://github.com/mozilla-services/consvc-shepherd/blob/58af2bf9ac766cffeb9f6d1e8049e955aebc56b9/Makefile#L26) some common PEP-8 conventions. I think we should stop ignoring those (in a separate PR of course, as we may need to change a bunch of things).

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "consvc_shepherd.settings")
+django.setup()
+
+from consvc_shepherd.models import BoostrDeal, BoostrDealProduct, BoostrProduct
+
+@dataclass(frozen=True)
+class BoostrSyncConfig():
```
cbguder, line 15: Nit: empty parentheses are not necessary

```suggestion
class BoostrSyncConfig:
```

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+    """Object to wrap up various environment configuration for the boostr snyc script"""
+    base_url: str
+    headers: dict[str, str]
+    log: logging.Logger
+
+class BoostrApiError(Exception):
+    '''Raise this error whenever we don't get a 200 status back from boostr'''
+    pass
+
+def sync_boostr_data()-> None:
```
cbguder, line 25:
```suggestion
def sync_boostr_data() -> None:
```

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+
+def get_campaign_type(product_full_name: str) -> str:
+    if "CPC" in product_full_name:
+        return "CPC"
+    if "CPM" in product_full_name:
+        return "CPM"
+    return "?"
+
+
+sync_boostr_data()
```
cbguder, line 143: Adding this condition will make it easier to unit test functions in this file

```suggestion
if __name__ == "__main__":
    sync_boostr_data()
```

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+class BoostrApiError(Exception):
+    '''Raise this error whenever we don't get a 200 status back from boostr'''
+    pass
+
+def sync_boostr_data()-> None:
+    logger = logging.getLogger("sync_boostr_data")
+    try:
+        config = setup_config(logger)
+        upsert_products(config)
+        upsert_deals(config)
```
cbguder, line 30: Suggestion: consider wrapping these into a single class to simplify, like:
```
class BoostrLoader:
  def __init__(self, ...):
    ...

  def load(self):
    self.authenticate()
    self.upsert_products()
    self.upsert_deals()

  def authenticate():
    self._token = ...

loader = BoostrLoader(base_url, email, password)
loader.load()
```

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+    env = environ.Env()
+    boostr_base_url = env("BOOSTR_BASE_URL")
+    email = env("BOOSTR_EMAIL")
+    password = env("BOOSTR_PASSWORD")
+    headers = {
+        "Accept": "application/vnd.boostr.public",
+        "Content-Type": "application/json"
+    }
+    token = authenticate(boostr_base_url, email, password, headers)
+    headers["Authorization"] = f"Bearer {token}"
```
cbguder, line 45: Suggestion: you can store these in a requests Session and reuse it:

```
s = requests.Session()
s.headers.update({...})

s.get(f"{config.base_url}/deals", params=params)
```

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+    token = user_token_response.json()
+    return token["jwt"]
+
+def upsert_deals(config: BoostrSyncConfig) -> None:
+    deals_params = {
+        "per": 300,
+        "page": 0,
+        "filter": "all",
+    }
+    while True:
```
cbguder, line 66: Might wanna add a max page limit just to be safe.

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+        deals_params["page"]+=1
+
+        deals_response = requests.get(f"{config.base_url}/deals", params=deals_params, headers=config.headers)
+        if deals_response.status_code != 200:
+            raise BoostrApiError(f"Bad response status from /api/deals: {deals_response}")
+        deals = deals_response.json()
+        config.log.info(f"Fetched {len(deals)} deals for page {deals_params['page']}")
+
+        # Paged through all available records and are getting an empty list back
+        if (len(deals) == 0):
```
cbguder, line 76:
```suggestion
        if len(deals) == 0:
```

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+            config.log.info(f"Done. Fetched all the deals in {deals_params['page']-1} pages")
+            break
+
+        closed_won_deals = filter(lambda d: d["stage_name"] == "Closed Won", deals)
```
cbguder, line 80: Does the Boostr API allow filtering by this in the request? Seems like that would be a win.

Also nit: I think `[d for d in deals if d["stage_name"] == "Closed Won"]` is more idiomatic Python for this

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+        for deal in closed_won_deals:
+            boostr_deal, ok = BoostrDeal.objects.update_or_create(
```
cbguder, line 82: Do we check these `ok`s?

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+    deals_params = {
+        "per": 300,
+        "page": 0,
+        "filter": "all",
+    }
+    while True:
+        deals_params["page"]+=1
+
+        deals_response = requests.get(f"{config.base_url}/deals", params=deals_params, headers=config.headers)
+        if deals_response.status_code != 200:
```
cbguder, line 70: FWIW there's `response.ok` and `response.raise_for_status()`, but they're both less strict than this check

consvc_shepherd/migrations/0008_boostrdeal_boostrproduct_boostrdealproduct_and_more.py
```diff
@@ -0,0 +1,96 @@
...
+                    models.BigAutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("boostr_id", models.IntegerField(unique=True)),
+                ("full_name", models.CharField()),
+                ("campaign_type", models.CharField()),
```
cbguder, line 52: Should this be an enum field?

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+                config.log.debug(f'Upserted {len(deal_product["deal_product_budgets"])} months of budget for product: {product.id} to deal: {deal_id}')
+
+def upsert_products(config: BoostrSyncConfig)-> None:
+    products_params = {
+        "per": 300,
+        "page": 1,
+        "filter": "all",
+    }
+    products_response = requests.get(f"{config.base_url}/products", params=products_params, headers=config.headers)
+    products = products_response.json()
```
cbguder, line 121: Probably want to do this after we check the status code

sync_boostr_data.py
```diff
@@ -0,0 +1,143 @@
...
+    pass
+
+def sync_boostr_data()-> None:
+    logger = logging.getLogger("sync_boostr_data")
+    try:
+        config = setup_config(logger)
+        upsert_products(config)
+        upsert_deals(config)
+    except BoostrApiError as e:
+        logger.error(e)
```
cbguder, line 32: Is it possible/straightforward to run this whole operation in a transaction and revert when there is an error?

sync_boostr_data.py
cbguder: This turned out to be way easier than I expected, so here's a PR: https://github.com/mozilla-services/consvc-shepherd/pull/216

sync_boostr_data.py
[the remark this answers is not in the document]
mashalifshin: Oh great, this looks like The Way. I definitely wanted to do something better than leave it hanging out in the root dir as a random script, but I was not having success with moving it into `/scripts`, so I'll give this approach a go

sync_boostr_data.py
mashalifshin: Woo, thank you for taking care of that!

consvc_shepherd/migrations/0008_boostrdeal_boostrproduct_boostrdealproduct_and_more.py
```diff
@@ -0,0 +1,96 @@
...
+                ),
+            ],
+        ),
+        migrations.AddField(
+            model_name="boostrdeal",
+            name="products",
+            field=models.ManyToManyField(
+                related_name="products",
+                through="consvc_shepherd.BoostrDealProduct",
+                to="consvc_shepherd.boostrproduct",
```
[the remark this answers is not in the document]
mashalifshin: I think this is getting generated automatically for the foreign key, not sure why the casing looks like this here...

consvc_shepherd/admin.py
```diff
@@ -217,3 +220,49 @@ def delete_queryset(self, request, queryset) -> None:
         """Delete given AllocationSetting entry."""
         super(AllocationSettingAdmin, self).delete_queryset(request, queryset)
         metrics.incr("allocation.delete")
+
+
+class BoostrDealProductInline(admin.StackedInline):
```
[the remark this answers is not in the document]
mashalifshin: Yeah, you can view a Deal in the admin UI and see the Deal Products inline

sync_boostr_data.py
mashalifshin: Yes! Awesome suggestion thank you. Things are looking much more organized with the script now.

sync_boostr_data.py
mashalifshin: 🔥

sync_boostr_data.py
mashalifshin: I don't think this option is available on the API, but I have my eye on pulling less of the deals at some future point based on timestamp filtering that the API _does_ offer. Right now since 2016 is the staging ground year, pulling only the deals from the past couple years won't work either...This is in my notes and I'll keep noodling on it.

sync_boostr_data.py
mashalifshin: Lol I wish it did but no.

Thanks for pointing out the more Pythonic way, that's really nice!

sync_boostr_data.py
mashalifshin: I need to think more deeply about this after I come back from break, I've added it to my todos...

sync_boostr_data.py
mashalifshin: Good eye! Fixed

sync_boostr_data.py
mashalifshin: Punted on making this change here, but now I know to use those for the future

consvc_shepherd/migrations/0008_boostrdeal_boostrproduct_boostrdealproduct_and_more.py
mashalifshin: Yes, yes it should ;-)

mashalifshin:
@cbguder thank you so much for all the helpful comments! I believe I've addressed all of them (BoostrLoader class, using the session, campaign type enum, unit tests). The only one I didn't address is  the "wrapping in transaction" comment, which I ran out of time for, it will require a bit more thought.

I also have one of the unit tests skipped currently, it's hanging when I add the last mock, but I'll revisit it when I return from vacation.

Other than those two issues, I think this is in much better shape now thanks to your feedback!

cbguder (approved):
Apologies for the massive delay in approving this. I think this looks great, thank you!

There are a few outstanding suggestions, but none of them are blockers for this to land. They can all be done in future PRs as you see fit.

consvc_shepherd/tests/test_sync_boostr.py
```diff
@@ -0,0 +1,572 @@
...
+    BoostrApiError,
+    BoostrDeal,
+    BoostrDealProduct,
+    BoostrLoader,
+    BoostrProduct,
+    get_campaign_type,
+)
+
+
+class MockResponse:
```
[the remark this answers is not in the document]
mashalifshin: Totally! I debated whether to bring in this new testing dep here, and ended up deciding to go with the simple mock object for funsies, but maybe I would have been better off with the lib. Next iteration that I get a chance with, I'll mess around with `requests-mock` and see how it is

mashalifshin:
I've merged the change as approved, and today I'm starting to think about the follow up PR with the outstanding issues. These are:

- A limit on the loop that fetches deals, so that it will end eventually no matter what when it's running as a schedule job. I'm thinking a new configuration variable which will default to something reasonable given the amount of deals we currently expect to get back from boostr
- Think more deeply about and implement something reasonable around @cbguder's question: "Is it possible/straightforward to run this whole operation in a transaction and revert when there is an error?".

mashalifshin:
Upper limit on fetching deals: https://github.com/mozilla-services/consvc-shepherd/pull/225

PR #216, opened 2024-07-29
Remove flake8 exceptions

cbguder:
## Description

In #215 I noted I think we should stop ignoring PEP-8 conventions, but it might require changing a lot of files.

It turned out to be way easier than I expected, so this drops all ignored rules, except increasing the max line length to 120 characters (the default 79 characters is too short for modern screens, and we surpass it on >150 lines of code already).t

## PR Review Checklist


mashalifshin (approved):
Nice, thank you!

PR #224, opened 2024-08-19
Ae 504 m 1 update models and boostr sync script to add cpc cpm rates

nanagb:
## References

JIRA: [AE-504](https://mozilla-hub.atlassian.net/browse/AE-504 )

## Description
Draft PR to review and comment on code changes for syncing CPM/CPM/Impression data and change in how deals are pulled in.

## PR Review Checklist

[AE-504]: https://mozilla-hub.atlassian.net/browse/AE-504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/models.py
```diff
@@ -199,3 +202,146 @@ def to_dict(self) -> dict[str, Any]:
...
+    boostr_deal: ForeignKey = models.ForeignKey(BoostrDeal, on_delete=models.CASCADE)
+    boostr_product: ForeignKey = models.ForeignKey(
+        BoostrProduct, on_delete=models.CASCADE
+    )
+    budget: IntegerField = models.IntegerField()
+    month: CharField = models.CharField()
+
+
+class BoostrDealMediaPlan(models.Model):
+    """TODO"""
```
mashalifshin, line 328: Should there be a migration in this PR that is for this model? And same question for the BoostrDealMediaPlanLineItem model...

Assuming these are new tables that we will have in our DB, I think I'm unclear and would like to understand better why you chose to add `BoostrDealMediaPlan` and `BoostrDealMediaPlanLineItem` to our schema here in Shepherd.

My thinking was that we didn't want to necessarily replicate the way data is organized/structured in the Boostr API, unless it is specifically useful to us in some particular way to do so. My goal is to simplify/reduce data and schema from their API into the very bare minimum that we need to drive our inventory dashboard views.

Back before I left, iirc I was thinking that the rate column would go on the existing BoostrDealProduct table (since that already represents the deal's monthly product and its budget). 

So yeah hoping you can help me understand better the intention with this schema. Lmk if I'm missing something about why we might want to model the media plans and their line items here in Shepherd. 

Happy to keep chatting in the comments here, or we can connect on a call tomorrow and talk through it if you prefer.

nanagb: Yep happy to talk more tomorrow but this was mainly as a result of the Boostr API design choices. There's no way to just get a media plan from a deal id. The only way I saw was to do it in multiple steps and saving the data from step 1 made it so we didn't have to call the API again for the next step.

I'm open to changing this if there's something I missed

sync_boostr_data.py
```diff
@@ -0,0 +1,313 @@
...
+    for product in products:
+        BoostrProduct.objects.update_or_create(
+            boostr_id=product["id"],
+            full_name=product["full_name"],
+            campaign_type=get_campaign_type(product["full_name"]),
+        )
+    config.log.info(f"Upserted {(len(products))} products")
+
+
+def upsert_media_plans(config: BoostrSyncConfig) -> None:
```
mashalifshin, line 157: Sorry if you're already on it and this comment is redundant! 

This file was removed in the latest version that's now on `main`. The boostr sync code now lives as an admin command in `consvc_shepherd/management/commands`, so you'll need to move this new functionality over there

nanagb: Ah I didn't catch that. I'll move my new code over.

consvc_shepherd/models.py
mashalifshin: Ugh you're right there is no way to use the API to get media plans by deal ID. So we'll have to go the other direction, from the media plans to the deals. And yeah their API is quite a PITA! 

(I bet you're totally right that their own product isn't built on it. I've wondered if it is a generated API ... )

Anyway! This is rough but here is how I think we can do it to avoid duplicating their schema into Shepherd here:

Add an `upsert_rates` function which will run another big loop to paginate through ALL the media plans from the `GET /api/media_plans` endpoint

For each media plan:
   * we will get back a `deal_id` in the response so we'll know what deal that media plan corresponds to. Hold the `deal_id` in memory
   * Call the boostr API again for this media plan, using the `GET /api/media_plans/{media_plan_id}/line_items` to fetch the its line items
   * Iterate through the media plan line items, for each line item
       1. Find the corresponding `BoostrDealProduct` that we've already saved. I think we should be able to do that using 
       * the `deal_id` we are holding in memory
       * the product information that comes back under the `product` key from this `/line_items` response 
       * the `start_date` and `end_date` that comes back from this `/line_items` response to identify the correct month for the `BoostrDealProduct`. Or maybe the `line_item_monthlies` array. I'm not sure which is the best one to use here, would need to play around more and see the data to decide which is correct/easiest. And sometimes they provide a month as a string, other times start and end date as timestamps, so it may be easier to use one or the other. Currently we represented the month on the `BoostrDealProduct` as a string, but it may also make it easier to adjust that to a date time...
       2. Upsert the rate to the correct `BoostrDealProduct` (`rate` is also in the `/line_items` response)

mashalifshin: Lmk how that sounds to you and happy to discuss more tomorrow!

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -153,43 +155,76 @@ def upsert_deals(self) -> None:
...
-                self.upsert_deal_products(boostr_deal)
+                # self.upsert_deal_products()
                 self.log.info(f"Upserted products and budgets for deal: {deal['id']}")
 
-    def upsert_deal_products(self, deal: BoostrDeal) -> None:
-        """Fetch the deal_products for a particular deal and store them in our DB with their monthly budgets"""
-        deal_products_response = self.session.get(
-            f"{self.base_url}/deals/{deal.boostr_id}/deal_products"
-        )
+    def upsert_deal_products(self) -> None:
```
mashalifshin, line 161: Let me know if I am understanding this change correctly, and please clarify if I'm missing something: 

Before, we were fetching the deal products via the nested route `GET /deals/{deal_id}/deal_products`, and using an in-memory `deal` object and querying for the relevant product, in order to save deal_products to our DB. 

Now, we're using the paginated, non-nested `GET /deal_products` endpoint, and we're querying for the relevant deal by id, and still also querying for relevant product, in order to save deal_products to our DB.

What are the advantages of the new approach? And I think I'm confused with the goal of this change and why we'd want to do this change as part of saving the cpc and cpm rates?

consvc_shepherd/models.py
mashalifshin: Did you get a chance to check out the above suggested approach?

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -153,43 +155,76 @@ def upsert_deals(self) -> None:
...
+        page = 0
+        deals_product_params = {
+            "per": "300",
+            "page": str(page),
+            "filter": "all",
+        }
+        while True:
+            page += 1
+            deals_product_params["page"] = str(page)
+            print(f"{self.base_url}/deal_products")
```
mashalifshin, line 172: Don't forget to remove this print statement before merging

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -153,43 +155,76 @@ def upsert_deals(self) -> None:
...
 
     def load(self):
         """Loader entry point"""
+        start_time = time.time()
         self.upsert_products()
         self.upsert_deals()
+        self.upsert_deal_products()
+        end_time = time.time()
+        elapsed_time = end_time - start_time
+        print(f"Sync took {elapsed_time:.4f} seconds to run")
```
mashalifshin, line 227: Let's use the logger (`self.logger`) here instead of the print statement

mashalifshin (commented):
I ran the script (after uncommenting `self.upsert_deal_products()`) and it's working well!

I left a couple small nitpicky comments, and I have some general questions for you about the refactor, and a suggested approach that removes the need for the new media plan tables -- curious if you think that'll work

consvc_shepherd/models.py
nanagb: Not yet, but it sounds like it will work. I'll be testing it out today

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -153,43 +155,76 @@ def upsert_deals(self) -> None:
                 )
                 self.log.debug(f"Upserted deal: {deal['id']}")
 
-                self.upsert_deal_products(boostr_deal)
+                # self.upsert_deal_products()
```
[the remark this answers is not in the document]
nanagb: yep, I'll return it back once we have resolved the rate limiting issues

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: Yes, you got it.

This saves on API calls because it's returning 300 results at a time. In the worst case, a sync would require hundreds of API calls.

I initially made this change because it made the process faster, but now its also needed to avoid maxing out out rate limit.

This change is just for dealproducts. I haven't added the sync for cpc/cpm to this file yet.

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: Ty done

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: Done

sync_boostr_data.py
```diff
@@ -0,0 +1,313 @@
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
nanagb: Ok.

I'm just keeping this around for the code I added to it. Till be sure to remove it once I'm done transferring to your new file and testing your idea for adding CPC/CPM without the extra tables.

Makefile
```diff
@@ -77,3 +78,19 @@ dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
 .PHONY: local-test
 local-test: $(INSTALL_STAMP)
 	docker compose -f docker-compose.test.yml up --abort-on-container-exit
+
```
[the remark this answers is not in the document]
nanagb: Thanks. I will expand out the names, I was using this locally and decided it may be helpful for the rest of the team.

VER allows you to pass in the migration version to roll back to eg

make remove-dm VER=0008

mashalifshin (commented):
Thanks for the replies and clarifications, all looks great so far! Looking forward to the next iteration 🪨 🔛

Makefile
```diff
@@ -14,31 +19,50 @@ $(INSTALL_STAMP): pyproject.toml poetry.lock
 	touch $(INSTALL_STAMP)
 
 .PHONY: isort
-isort: $(INSTALL_STAMP)  ##  Run isort
+isort: $(INSTALL_STAMP)  ##  Run isort in --check-only mode
+	@echo "Running isort..."
 	$(POETRY) run isort --check-only $(APP_DIRS) --profile black
 
+.PHONY: isort-fix
+isort-fix: $(INSTALL_STAMP)  ##  Run isort to fix imports
```
mashalifshin, line 27: 💖 Love all these convenient lint fixing ones, so useful! Been just pulling them out of my shell history and this is better!

Makefile
```diff
@@ -75,5 +99,21 @@ dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
 	docker compose up
 
 .PHONY: local-test
-local-test: $(INSTALL_STAMP)
+local-test: $(INSTALL_STAMP)  ##  local test
 	docker compose -f docker-compose.test.yml up --abort-on-container-exit
+
+.PHONY: d-makemigrations-empty
+d-makemigrations-empty: ##  Run create an empty migrations file via the docker container
```
mashalifshin, line 106: These also are fantastic! 

I don't know how others feel, but I'm always running this in docker so I wouldn't mind having these get the 'default' naming like `migrate`, `makemigrations`, etc. And then we'd change the existing `migrate` & friends to `local-migrate`, giving those the `local-` prefix that we already use here, for example in `local-test`.

Just an idea though, totally not opinionated on that!

Makefile
```diff
@@ -2,10 +2,15 @@ APP_DIRS := consvc_shepherd contile openidc schema
 COV_FAIL_UNDER := 95
 INSTALL_STAMP := .install.stamp
 POETRY := $(shell command -v poetry 2> /dev/null)
+VER :=
```
mashalifshin, line 5: @nanagb if you'd like to (and want a quick win after wrestling with the darn Boostr API), you could pull these `Makefile` updates out into their own separate PR so we could merge it fast and the whole team can start making use of them. 

But also totally fine to keep them here and wait til this larger change lands, I'm just excited about 'em :-)

mashalifshin (commented):
Loving the `Makefile` updates 🔥

Makefile
nanagb: Makes sense, agree. I would like that switch too!

Makefile
nanagb: Sounds like a nice way to end the week :D

PR #225, opened 2024-08-23
[AE-532] Add upper limit to boostr deals fetching

mashalifshin:
## References

JIRA: [AE-532](https://mozilla-hub.atlassian.net/browse/AE-532)

@cbguder [made a good point](https://github.com/mozilla-services/consvc-shepherd/pull/215#discussion_r1702065723) on the last iteration of the Boostr sync script, that the `while True` loop is dangerous and might not ever finish if we receive something unexpected back from the API.

This change addresses that by adding an upper limit of how many times we fetch deals pages. It is set up for a default max of `50`, and can be adjusted by passing an optional parameter to the script.

## PR Review Checklist

[AE-532]: https://mozilla-hub.atlassian.net/browse/AE-532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/consvc_shepherd/tests/test_sync_boostr.py
+++ b/consvc_shepherd/tests/test_sync_boostr.py
@@ -476,6 +476,21 @@ class TestSyncBoostrData(TestCase):
     @mock.patch("requests.post", side_effect=mock_post_success)
     @mock.patch.object(requests.Session, "get", side_effect=mock_get_success)
     @mock.patch(
+        "consvc_shepherd.models.BoostrDeal.objects.update_or_create",
+        side_effect=mock_update_or_create_deal,
+    )
+    @mock.patch.object(BoostrLoader, "upsert_deal_products")
+    def test_upsert_deals_respects_max_deal_pages_limit(
+        self, mock_upsert_deal_products, mock_update_or_create, mock_get, mock_post
+    ):
+        """Test that upsert_deals respects the max_deal_pages limit"""
+        loader = BoostrLoader(BASE_URL, EMAIL, PASSWORD, 3)
+        loader.upsert_deals()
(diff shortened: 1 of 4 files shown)


consvc_shepherd/tests/test_sync_boostr.py
```diff
@@ -473,6 +473,21 @@ def test_upsert_deals_fail(self, mock_get, mock_post):
...
+        side_effect=mock_update_or_create_deal,
+    )
+    @mock.patch.object(BoostrLoader, "upsert_deal_products")
+    def test_upsert_deals_respects_max_deal_pages_limit(
+        self, mock_upsert_deal_products, mock_update_or_create, mock_get, mock_post
+    ):
+        """Test that upsert_deals respects the max_deal_pages limit"""
+        loader = BoostrLoader(BASE_URL, EMAIL, PASSWORD, 3)
+        loader.upsert_deals()
+        assert 3 == mock_get.call_count
```
mashalifshin, line 489: Ran into [something interesting here](https://docs.python.org/3.5/library/unittest.mock-examples.html#coping-with-mutable-arguments) in trying to assert that each call of the `mock_get` had the correct page parameters.

Because the outer loop around that mock call mutates the request params variable, and because mocks pass their arguments by reference, I couldn't assert on the specific page params without doing some more work to capture the params as they change in each run of the loop.

I decided this assertion was sufficient to verify that the max_deal_page_limit functionality works, but if the reviewer feels differently I'd be happy to add params checks.

nanagb (approved):

PR #227, opened 2024-08-23
AE-538 makefile improvements

nanagb:
## References

JIRA: [AE-538](https://mozilla-hub.atlassian.net/browse/AE-538)

## Description

Added a working help target and multiple new commands for migrations. This is also a first step in getting the Shepherd Makefile in alignment with the MARS file

## PR Review Checklist

[AE-538]: https://mozilla-hub.atlassian.net/browse/AE-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,14 @@ APP_DIRS := consvc_shepherd contile openidc schema
 COV_FAIL_UNDER := 95
 INSTALL_STAMP := .install.stamp
 POETRY := $(shell command -v poetry 2> /dev/null)
+VER :=
 
 # This will be run if no target is provided
 .DEFAULT_GOAL := help
+
+.PHONY: help
+help: ##  show this help message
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,5 +99,21 @@ dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
 	docker compose up
 
 .PHONY: local-test
-local-test: $(INSTALL_STAMP)
+local-test: $(INSTALL_STAMP)  ##  local test
```
mashalifshin, line 102: Whoops, reading this change made me realize I had the wrong idea of the `local-` prefix here, and this actually also runs in docker, so local here just means "not in CI". But I still like the changes here and think the naming makes decent sense

mashalifshin (approved):
These are so helpful, thank you for adding them! I ran each one to spot-check and they are all working great. :shipit:  

One thing I'm wondering is if we might like a `make up` command or something like that for `docker compose up --build`? 

And then also thinking about those last `make` commands that use `docker exec` and wondering if we could then chain them with `make up` so that they are self-contained and spin up the containers on their own if needed. This could take some figuring out though, not sure if we'd need to check if one is already spun up first? 

That can be in the next iteration, though, as this PR LGTM and is already great to merge as-is!

PR #229, opened 2024-08-26
unified api preview using wrong placement names

dmueller:
## References

JIRA: [DISCO-TODO](https://mozilla-hub.atlassian.net/browse/DISCO-TODO)

## PR Review Checklist

### DIFF
--- a/consvc_shepherd/preview.py
+++ b/consvc_shepherd/preview.py
@@ -322,18 +322,25 @@ def get_unified(env: Environment, country: str, is_mobile: bool = False) -> Ads:
     # placement names will vary for preview and experiment environments, whereas
     # dev & prod have the same placements served by different kevel networks
     spocs_placement = "newtab_spocs"
-    tiles_placement = "newtab_tiles"
+    tile_one_placement = "newtab_tile_1"
+    tile_two_placement = "newtab_tile_2"
+    tile_three_placement = "newtab_tile_3"
 
     # load spocs & tiles, then map them to the same shape
     body = {
         "user_context_id": f"{user_context_id}",  # UUID -> str
         "placements": [
             {"placement": spocs_placement, "count": 10},
-            {"placement": tiles_placement, "count": 3},
+            {"placement": tile_one_placement, "count": 1},
+            {"placement": tile_two_placement, "count": 1},
+            {"placement": tile_three_placement, "count": 1},
         ],
     }
 
     r = requests.post(f"{env.mars_url}/v1/ads", json=body, timeout=30)
+    r_json = r.json()
+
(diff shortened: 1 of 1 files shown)


mashalifshin (approved):
Thanks so much for noticing this was broken and fixing this! 🙌

mashalifshin:
@dmueller what kinds of automated testing do you think we could add here to catch issues like this in the future?

dmueller:
> @dmueller what kinds of automated testing do you think we could add here to catch issues like this in the future?

I attribute this to be a by-product of the unified work being fluid right now since we don't have any prod traffic yet. In reality MARS shouldn't remove any placement names that were ever used and instead have those result in an empty response. Now that kind of thing would still have broken shepherd in that tiles wouldn't have rendered anymore, but at least it wouldn't have also broken spocs.

Otherwise that adjustment could be an easy breaking change for clients like this, which is a bad experience for customers.

If we had automated tests for this, I think they'd essentially be a second list of placement names that should have a successful request to help us avoid breaking changes

mashalifshin:
> > @dmueller what kinds of automated testing do you think we could add here to catch issues like this in the future?
> 
> I attribute this to be a by-product of the unified work being fluid right now since we don't have any prod traffic yet. In reality MARS shouldn't remove any placement names that were ever used and instead have those result in an empty response. Now that kind of thing would still have broken shepherd in that tiles wouldn't have rendered anymore, but at least it wouldn't have also broken spocs.
> 

Sorry for the slow reply and thanks so much for this answer, that totally makes sense. Keep forgetting Unified API isn't quite prod-ready yet!

> Otherwise that adjustment could be an easy breaking change for clients like this, which is a bad experience for customers.
> 
> If we had automated tests for this, I think they'd essentially be a second list of placement names that should have a successful request to help us avoid breaking changes

Yeah it feels like something best verified by tests on MARS itself and maybe in the contract tests -- haven't taken a look at those yet, but seems reasonable that they would verify all supported placement names

mashalifshin:
> > @dmueller what kinds of automated testing do you think we could add here to catch issues like this in the future?
> 
> I attribute this to be a by-product of the unified work being fluid right now since we don't have any prod traffic yet. In reality MARS shouldn't remove any placement names that were ever used and instead have those result in an empty response. Now that kind of thing would still have broken shepherd in that tiles wouldn't have rendered anymore, but at least it wouldn't have also broken spocs.
> 
> Otherwise that adjustment could be an easy breaking change for clients like this, which is a bad experience for customers.
> 
> If we had automated tests for this, I think they'd essentially be a second list of placement names that should have a successful request to help us avoid breaking changes

Oh right, I keep forgetting y'all are still free to move quickly on MARS since it's not prod-facing yet! Thanks for the explanation on that. It also sounds like this could be a good thing to verify in the QA contract tests, if not already ...

PR #232, opened 2024-08-29
[AE-534] Respect the rate limits

mashalifshin:
## References

JIRA: [AE-534](https://mozilla-hub.atlassian.net/browse/AE-534)

## Description
Recently, Boostr surprised us by [introducing rate limits to the API](https://help.boostr.com/en/articles/9006378-public-api-rate-limits-and-best-practices). This simple change:

- Refactors the script to have a wrapper for Boostr Api requests
- Slows down requests by an interval that's just slower than what the Boostr rate limit allows. This keeps us under their rate limits and not needing to handle 429s, but we can add that in a subsequent PR if it becomes necessary

I'm thinking this will unlock the ability to run the script in all environments when @cyptm's scheduled job changes land, even though it runs slower, that should be fine.

Longer term, we can look at other ways to limit the amount of data we are pulling by creating a real staging environment (AE-501) and by further enhancing the script to only pull what has changed since the last time (no ticket yet as it would be an enhancement and we likely have bigger fish to fry)

To verify this change locally, run the script and see that it chugs along slowly but surely staying under the rate limits and not receiving 429s. I'll do more local testing as well. 

## PR Review Checklist

- [x] The PR title starts with the JIRA issue reference, format `[AE-534]]`, and has the same title (if applicable).

[AE-534]: https://mozilla-hub.atlassian.net/browse/AE-534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AE-534]: https://mozilla-hub.atlassian.net/browse/AE-534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/consvc_shepherd/tests/test_sync_boostr.py
+++ b/consvc_shepherd/tests/test_sync_boostr.py
@@ -2,10 +2,10 @@
 
 from unittest import mock
 
-import requests
 from django.test import TestCase, override_settings
 
 from consvc_shepherd.management.commands.sync_boostr_data import (
+    BoostrApi,
     BoostrApiError,
     BoostrDeal,
     BoostrDealProduct,
@@ -21,18 +21,19 @@ class MockResponse:
     def __init__(self, json_data: object, status_code: int):
         self.json_data = json_data
         self.status_code = status_code
+        self.ok = 200 <= self.status_code < 400
 
     def json(self):
         """Mock json data"""
         return self.json_data
 
 
-def mock_post_success(*args, **kwargs) -> MockResponse:
+def mock_post_token_success(*args, **kwargs) -> MockResponse:
     """Mock successful POST /user_token requests to boostr"""
     return MockResponse({"jwt": "i.am.jwt"}, 201)
 
 
-def mock_post_fail(*args, **kwargs) -> MockResponse:
+def mock_post_token_fail(*args, **kwargs) -> MockResponse:
     """Mock failed POST /user_token requests to boostr"""
     return MockResponse({"uh": "oh"}, 401)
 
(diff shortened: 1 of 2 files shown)


consvc_shepherd/tests/test_sync_boostr.py
```diff
@@ -369,37 +373,39 @@ def mock_update_or_create_deal_product(
 class TestSyncBoostrData(TestCase):
```
mashalifshin, line 373: I'd like to do some more re-organizing and cleaning up this test file as it grows large and unwieldy, but I'm holding off to limit the scope on this PR since I know others are working in this file. Specifically I'd like to:
- split tests off into multiple test suites instead of single `TestSnycBoostrData`
- leverage some class-level `mock.patch`ing to DRY up the decorators
- pull out the example responses from Boostr into separate fixtures files to cut down on setup boilerplate in the test file
- break up some of the functions in snyc_boostr_data into smaller components so they are more easy to test in smaller, more single-purpose test cases

nanagb (approved):

mashalifshin:
@nanagb noted:

It still got hit with rate limits and broke (errored). I think the next iteration needs to retry on a 429 error instead of existing. Because the rate limit is temp. Also  look at the retry after  header because in my tests, after that wait time, we are able to hit the api at full speed again. I lost track of how long the sync took before the error happened but it was over 10 mins.

PR #240, opened 2024-09-07
Set default request rate to 1 per second for the boostr sync script

mashalifshin:
## Problem Statement

I am not sure that the current request rate setting will allow the script to do a full sync given the Boostr API's rate limits, based on some quick tests I've run. As we're preparing to merge our scheduled job to run this script in all environments on a regular basis, I thought it would be helpful to be able to configure this value. 

## Proposed Changes

* Set up a parameter for the request rate instead of hard coding it, meaning this value can be configurable without making a PR and pushing a code change here.
* Set the default value of requests every 1 seconds for the script. At this default setting, I have managed a successful full sync of the current boostr data in 45 minutes.
* A few additional documentation updates, fixes, and helpful tips

## Verification Steps

The specs verify that the Boostr api wrapper code does in fact wait between requests, and that a specified interval is used by the wrapper. Let me know if you see any other opportunities for adding tests.

For manual verification, you can run the script locally (see `syncBoostrData.md` for usage details) and pass a very high value to see if the script is respecting the value you passed.

### DIFF
--- a/consvc_shepherd/management/commands/sync_boostr_data.py
+++ b/consvc_shepherd/management/commands/sync_boostr_data.py
@@ -13,7 +13,7 @@ from django.core.management.base import BaseCommand
 from consvc_shepherd.models import BoostrDeal, BoostrDealProduct, BoostrProduct
 
 MAX_DEAL_PAGES_DEFAULT = 50
-RATE_LIMIT_REQUEST_INTERVAL_SECS = 0.7
+REQUEST_INTERVAL_SECONDS_DEFAULT = 2
 
 
 class Command(BaseCommand):
@@ -26,7 +26,7 @@ class Command(BaseCommand):
         parser.add_argument(
             "base_url",
             type=str,
-            help="The base url for the Boostr API, eg. https://app.boostr.com/api/",
+            help="The base url for the Boostr API, eg. https://app.boostr.com/api",
         )
         parser.add_argument(
             "--max-deal-pages",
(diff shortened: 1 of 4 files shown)


consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -36,6 +40,14 @@ def add_arguments(self, parser):
                 {MAX_DEAL_PAGES_DEFAULT} pages. Currently we have ~14 pages of deals in Boostr, so this default max
                 should be sufficient for a while.""",
         )
+        parser.add_argument(
```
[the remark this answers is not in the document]
mashalifshin: Yeah, as you saw I ended up going with an optional command line arg for this, following the logic of putting secrets in the env, and optional values as args.

I tried to do it so your scheduled job PR won't require any changes after this. Just got a sync to succeed with an interval of 1 second, and it runs in 45 minutes (unlike last Friday's debacle), so I think this should get us the data in all environments when your PR is merged

docs/operations/syncBoostrData.md
```diff
@@ -34,7 +35,21 @@ upper limit on deal pages.
...
 Usage:
 ```sh
-python manage.py sync_boostr_data
+python manage.py sync_boostr_data --max-deal-pages 15
+```
+
+#### --request-interval-seconds
+The script can take an optional named argument, `--request-interval-seconds`, which
+controls the rate of requests to the Boostr API in order to stay under their rate limits.
+By default, the sync code will wait 2 seconds between requests.
```
nanagb, line 44: Based on the rest of your code and comment, I think this should say 1 second

nanagb (approved):

docs/operations/syncBoostrData.md
mashalifshin: Good eye, thank you!

PR #241, opened 2024-09-09
[AE-565] Fetch and save the country for products after creating a new country field in the BoostrProduct table

akhan-mozilla:
## References

JIRA: [AE-565](https://mozilla-hub.atlassian.net/browse/AE-565)

## Description
1. Added a new country column to the BoostrProducts table.
2. Added a function to the Boostr script file to fetch and save the country code in the table, based on possible country codes that we might receive from the Boostr API.

## PR Review Checklist

[AE-565]: https://mozilla-hub.atlassian.net/browse/AE-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/migrations/0010_boostrproduct_country.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("consvc_shepherd", "0009_alter_boostrproduct_campaign_type"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="boostrproduct",
+            name="country",
+            field=models.CharField(blank=True),
```
nanagb, line 16: We can use [choices](https://docs.djangoproject.com/en/5.1/ref/models/fields/#choices) here

nanagb (changes requested):

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -232,3 +235,15 @@ def get_campaign_type(product_full_name: str) -> str:
         return BoostrProduct.CampaignType.FLAT_FEE
     else:
         return BoostrProduct.CampaignType.NONE
+
+
+def get_country(product_full_name: str) -> str:
+    """Return the country code found in the product's full name."""
+    country_codes = ["US", "UK", "DE", "FR", "CA", "IT", "SP"]
```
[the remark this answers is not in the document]
akhan-mozilla: Understood. For SP, it appears to also be used for Spain. I will keep it in the list for now; however, I will update the list once confirmed by the Ad Ops team.

consvc_shepherd/migrations/0010_boostrproduct_country.py
akhan-mozilla: Thanks for pointing this out. I've made the change.

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -232,3 +235,30 @@ def get_campaign_type(product_full_name: str) -> str:
...
+        "IT",
+        "PL",
+        "AT",
+        "NL",
+        "LU",
+        "CH",
+        "BE",
+        "SP",
+    ]
+    country_found = "ES"
```
nanagb, line 258: `country_found = "ES"` Will still save ES as the country if no match is found in the list. Use `    country_found = ""`

consvc_shepherd/models.py
nanagb, line 1: I would like to see something on our end that standardizes  how incorrectly entered country codes in Boostr is stored in our shepherd db. This will ensure that when we do country breakdowns for reporting, it's accurate and not split into two. eg UK and GB. SP and ES. etc

nanagb (approved):

PR #242, opened 2024-09-11
[SHEP-13] Add CampaignOverview model and its summary to Django admin with filters

akhan-mozilla:
## References

[SHEP-13](https://mozilla-hub.atlassian.net/browse/SHEP-13)

## Problem Statement
Created the CampaignOverview model and displayed the campaign summary, broken down by month, country, placement, and advertiser.

## Proposed Changes

1. Added the CampaignOverview model to the database
2. Displayed the CampaignOverview and its summary in the Django admin
3. Applied filters for month, placement, country, and advertiser to the summary

## Verification Steps

1. Run the migrations to create the table.
2. Add test data to the CampaignOverview model to view the summary of campaigns in relation to deals and advertisers.

## Visuals

If applicable, provide screenshots or videos here with the expected new behavior. If not, please remove this section.

[SHEP-12]: https://mozilla-hub.atlassian.net/browse/SHEP-12?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SHEP-13]: https://mozilla-hub.atlassian.net/browse/SHEP-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/models.py
```diff
@@ -339,3 +340,72 @@ class BoostrDealProduct(models.Model):
...
+    seller : CharField
+        Seller
+    created_on : DateTimeField
+        Date of deal record creation (shepherd DB timestamp metadata, not boostr's)
+    updated_on : DateTimeField
+        Date of deal record update (shepherd DB timestamp metadata, not boostr's)
+
+    Methods
+    -------
+    __str__(self)
```
nanagb, line 371: the __str__ method is missing

consvc_shepherd/models.py
```diff
@@ -339,3 +340,72 @@ class BoostrDealProduct(models.Model):
...
+        Return the string representation for a Boostr Campaign
+
+    """
+
+    ad_ops_person: CharField = models.CharField()
+    notes: CharField = models.CharField()
+    kevel_flight_id: IntegerField = models.IntegerField()
+    net_spend: IntegerField = models.IntegerField()
```
nanagb, line 379: The spend values are not always whole numbers. I've been using Decimal for monetary values. I read it's more accurate for calculations.

nanagb (changes requested):

consvc_shepherd/models.py
akhan-mozilla: Sure, I just updated the field type to decimal.

consvc_shepherd/tests/test_models.py
```diff
@@ -49,3 +57,75 @@ def test_to_dict_produces_correctly(self) -> None:
...
+            impressions_sold=500,
+            seller="Test Seller",
+            deal=self.deal,
+        )
+        campaign.save()
+        self.assertEqual(campaign.net_ecpm, (2000 / 500) * 1000)  # 4000
+
+    def test_str_method(self):
+        """Verify that the __str__ method returns the correct string representation."""
+        expected_str = "John Doe - 12345"
```
nanagb, line 117: Expected should be: Campaign 12345 - John Doe
`"Campaign {self.kevel_flight_id} - {self.ad_ops_person}"`

akhan-mozilla: Yes, I've already added fix commit for this.

nanagb: Awesome you beat me to it :D thanks!

consvc_shepherd/admin.py
```diff
@@ -268,3 +272,204 @@ class BoostrProductAdmin(admin.ModelAdmin):
...
+        aggregated_data = (
+            aggregated_query.values("advertiser")  # Group by advertiser
+            .annotate(
+                total_impressions_sold=Sum(
+                    "campaignoverview__impressions_sold", output_field=FloatField()
+                ),
+                total_net_spend=Sum(
+                    "campaignoverview__net_spend", output_field=FloatField()
+                ),
+                latest_note=Subquery(latest_notes_subquery),
```
nanagb, line 449: We may want  to let the user know they are looking at the latest note,seller,ad ops rep.

I thought it was a bug until I looked at the code. 😄 

Other options to consider:
1. Group by, advertiser, note, seller,ad ops
2. Display  something like "..." when the contents of an aggregation are not all the same
3. Don't show the 3 columns on the summary (note, seller,ad ops)

nanagb (approved):

consvc_shepherd/migrations/0011_campaignoverview_campaignoverviewsummary.py
```diff
@@ -0,0 +1,60 @@
+# Generated by Django 4.2.14 on 2024-09-11 19:56
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
mashalifshin, line 7: I think we want to split this CampaignOverview model into two separate concepts.

1. A Campaign model, backend by a DB table called Campaigns. The fields should be ad_ops_person, notes, kevel_flight_id, and a foreign key to the Deals table since Deals have many Campaigns
3. A CampaignOverview model, backed by a database view. This shouldn't have any of its own fields, but should be composed of existing fields from Deals, and Campaigns. For ad_ops_person, kevel_flight_id, and notes, we pull from Campaigns. For Net Spend, Impressions sold, Net eCPM, and Seller, we pull from the Deals table and calculate the result in the view.

mashalifshin (changes requested):
@akhan-mozilla This is an awesome start!! I am requested some changes around teh schema design before merging this PR. Details in a comment on the migration file.

consvc_shepherd/migrations/0011_campaignoverview_campaignoverviewsummary.py
mashalifshin: Just to add on a bit more re. how this will fit in with the frontend, my thinking is we'll have an endpoint like `GET /campaigns_overview` in our dashboard API, and that endpoint will query CampaignsOverview for the data it needs to show the table.

(This is definitely stretching the definitions of RESTful API but I'm not personally very particular about conforming very strictly as long as it's clear what is happening, I still like the idea of using a RESTful API for all the standardization/tooling we get out of the box. But open to other opinions)

akhan-mozilla: > A Campaign model, backend by a DB table called Campaigns. The fields should be ad_ops_person, notes, kevel_flight_id, and a foreign key to the Deals table since Deals have many Campaigns

Sure, I'll make that change. I initially considered the relationship between deals and campaigns as one-to-many (one deal to many campaigns). Based on your suggestion, we should establish this one-to-many relationship, where campaigns are linked to deals, as the deals table already contains multiple campaigns for each advertiser. This way, we don't need to duplicate the fields from the deals table in the campaigns model and can reference them directly.

consvc_shepherd/admin.py
akhan-mozilla: Thanks for pointing this out. I'll remove the three columns (notes, seller, and ad ops) from the campaign summary. Additionally, we need to display the campaign summary grouped by advertisers, which I've already added.

consvc_shepherd/migrations/0011_campaignoverview_campaignoverviewsummary.py
mashalifshin: Thanks so much for including the diagram, as well as the example data, that's super helpful here! 

I think with seeing this laid out so clearly, I now understand why in your initial schema design you wanted to put the impressions and the net spend on the `Campaigns` table as well. I wonder also if the `Campaigns` need their own start dates and end dates, too... This made me realize that I need a better understanding of how `Campaigns` relate to `Deals` when there are multiple child `Campaigns` per `Deal`.

I think we should watch [Laura's Kevel video](https://mozilla.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=ec468d57-307f-4ecb-ab94-b1c601784122) and/or check with Laura to be able to model this correctly.

-----

Let me also try out the idea of adding those same fields to `Campaigns` that are on `Deals`, with the thought that child `Campaigns` fields should "add up to" the parent `Deal`'s totals for impressions and amount. 

* When a `Deal` has a single `Campaign`, the parent `Deal` and the child `Campaign` will have identical net spend, impressions, and start and end dates

* When we have multiple `Campaigns` belonging to a `Deal`, it seems like we may want to support per-`Campaign` fields that total up to the values in the parent `Deal`. Trying that idea out with an example here:

```
Campaign schema: ad_ops_person, notes, kevel_flight_id, deal_id, amount, impressions, start_date, end_date
```


```
Deal: (advertiser, $5000, 1000000, 50, seller_name, 10/1/2024, 10/31/2024)
```

```
child Campaign: ('John Doe', 'Campaign for new product launch', 'KF123-A', 1, $3000, 600000, 10/1/2024, 10/15/2024)
child Campaign: ('Alice Brown', 'Initial brand awareness campaign', 'KF123-B', 1, $2000, 400000, 10/16/2025, 10/31/2024)
```

* the amounts on the child `Campaigns` add up to the amount in the parent `Deal` (`$3000 + $2000 = $5000`) 
* the impressions of the child `Campaigns` add up to the impressions in the parent `Deal` (`600000 + 400000 = 1000000`)
* the date ranges of the child `Campaigns` are within the start dates of the parent Deal (I'm guessing Campaign date ranges can overlap but maybe not? Another question for Laura)

akhan-mozilla: Yes, that's correct, While fetching data from the Boostr API, we should automatically create a child campaign and add the amount, impressions, start_date, and end_date to the campaigns table along with the deal ID. This way, the campaign can be further divided into multiple campaigns for that deal, as you explained in your example above.

akhan-mozilla: @mashalifshin   I've watched Laura's Kevel video, and it's clear that we need to store the `net_spend`, `impressions`, `start_date`, and `end_date` fields in the Campaigns model as these will be associated with campaigns. When displaying Deals, we should calculate the sum of `net_spend` and `impressions` from the related campaigns and show them in the Deals table.

I'll update the schema to include `start_date` and `end_date` in the Campaigns model as well. I'll keep these fields in the Deals table for now, but in the future, they can be removed and referenced directly from the Campaigns model.

mashalifshin: Awesome, so glad that was helpful and gave clarity on this decision! Agreed re. adding all those fields to the `Campaign` model.

And also agreed on keeping those fields on the `Deal` as well, at least for now while we think about it a bit more. I see that as being different, also useful information. It's really the `BoostrDeal` model, which we've been shortening to `Deal` for this discussion, but the full model name is there to help us remember that this represents the impressions and spend amount that was agreed by Sales team. The data in `Campaign` model is more about how we plan to deliver on that deal.

So if we keep that info on the `BoostrDeal` table, we could for example 
- support helpful validations like making sure that the total amount of all the child `Campaign`s doesn't except the total specified in the `BoostrDeal`
- show helpful info about how much spend or impressions remain from the parent Deal given the existing child `Campaign`s

akhan-mozilla: Yes, that’s correct. I'll add those validations while working on the campaigns CRUD(SHEP-12) to ensure that the total amount of child campaigns matches their parent deals, and the same will be done for impressions.

I've updated the campaigns model by adding start_date and end_date, and I’ve also added the campaigns summary view.

consvc_shepherd/tests/factories.py
```diff
@@ -26,8 +26,16 @@ class UserFactory(factory.django.DjangoModelFactory):
...
         """Meta class to bind model to get_user_model()."""
 
         model = get_user_model()
+
+
+class AdminUserFactory(UserFactory):
+    """Factory for creating admin user instances with staff and superuser privileges."""
+
+    is_staff: bool = True
+    is_superuser: bool = True
```
kauana, line 41: Hey @akhan-mozilla, could you please explain why we added the `AdminUserFactory`? Thank you

akhan-mozilla: I've added tests for campaign filters through their URLs, which require specific permissions. This setup is important for accessing the admin URLs during testing without encountering unauthorized errors. Additionally, this approach will also be used for other routes, as it applies not only to campaigns.

consvc_shepherd/migrations/0012_campaign.py
```diff
@@ -0,0 +1,50 @@
...
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("ad_ops_person", models.CharField()),
+                ("notes", models.CharField()),
+                ("kevel_flight_id", models.IntegerField()),
+                ("net_spend", models.DecimalField(decimal_places=2, max_digits=12)),
+                ("impressions_sold", models.IntegerField()),
+                ("net_ecpm", models.FloatField()),
```
mashalifshin, line 31: Took a quick glance and these changes are looking good! I'll take another detailed look on Monday, but in the meantime noticed one more adjustment I think we need here.

I believe `net_ecpm` should not be a first-order field on the Campaign, but is instead calculated from the `net_spend` and `impressions_sold` field. 

Please let me know if I'm off base here, I'm going off [the dashboard MVP doc](https://docs.google.com/document/d/1_M9uvN7n0czOju8CQxYFgOjf3VVqGeHCLUffcXLlwgM/edit#heading=h.k2q5uhbb8t8h) and I still haven't had a chance to view Laura's Kevel video.

akhan-mozilla: Yes, the `net_ecpm` is being calculated based on `impressions_sold` and `net_spend`. I've removed their fields from the models and ensured that the calculated value is being returned.

mashalifshin: Great, thank you!

nanagb:
> @akhan-mozilla The schema looks great!
> 
> It looks like you will need to re-create your migrations, since the ones you have on this branch conflict with what is on `main`. This process is very do-able but a bit tricky, happy to pair with you on this or lmk if I can help in any way. Here's what to do.
> 
>     1. On your branch, you need to roll back the DB schema to before your branch diverged with main, so roll back to migration 0010 with `python manage.py migrate consvc_shepherd 0010`
> 
>     2. On your branch, you're going to delete all the migration files you created, everything that's after `0010`.
> 
>     3. Get your branch up to date with `main` and pull in new migrations from there
> 
>     4. Migrate your DB with the new migrations from `main`
> 
>     5. Run `makemigrations` again to create your same migration files anew. Now they will start from the migrations that are on `main`
> 
>     6. Migrate your DB

I added a new make command that should do all this in one step. 
`make remove-migration-fix VER=0010`

akhan-mozilla:
> @akhan-mozilla The schema looks great!
> 
> It looks like you will need to re-create your migrations, since the ones you have on this branch conflict with what is on `main`. This process is very do-able but a bit tricky, happy to pair with you on this or lmk if I can help in any way. Here's what to do.
> 
> 1. On your branch, you need to roll back the DB schema to before your branch diverged with main, so roll back to migration 0010 with `python manage.py migrate consvc_shepherd 0010`
> 2. On your branch, you're going to delete all the migration files you created, everything that's after `0010`.
> 3. Get your branch up to date with `main` and pull in new migrations from there
> 4. Migrate your DB with the new migrations from `main`
> 5. Run `makemigrations` again to create your same migration files anew. Now they will start from the migrations that are on `main`
> 6. Migrate your DB

Thanks, @mashalifshin  and @nanagb , for letting me know. I’ve just updated the migrations following these steps.

mashalifshin:
@akhan-mozilla There are still issues with the migrations on this branch. Can you please take another try at fixing them up? I am also happy to pair on it and/or try to fix myself on my local env (I thought my instructions would work but maybe I missed something?)

Where we want to end up is that the migrations that are currently on `main` are not modified on this branch ('cause unlike the rest of the code, the migrations represent the state of the deployed DBs, so we can't modify those in place, we can only "go forward" by adding more new files). 

So that means `0011`, and `0012` migrations need to remain un-changed (they should not be in this diff), and then your new schema can start at `0013`. Right now it looks like your new schema got into the same migration file as some existing sync status stuff ...

akhan-mozilla:
> @akhan-mozilla There are still issues with the migrations on this branch. Can you please take another try at fixing them up? I am also happy to pair on it and/or try to fix myself on my local env (I thought my instructions would work but maybe I missed something?)
> 
> Where we want to end up is that the migrations that are currently on `main` are not modified on this branch ('cause unlike the rest of the code, the migrations represent the state of the deployed DBs, so we can't modify those in place, we can only "go forward" by adding more new files).
> 
> So that means `0011`, and `0012` migrations need to remain un-changed (they should not be in this diff), and then your new schema can start at `0013`. Right now it looks like your new schema got into the same migration file as some existing sync status stuff ...

Now  I ensured that the existing migrations from the main branch are not replaced/removed and that my changes create a new migration. Please check if this resolves the conflicts.

mashalifshin (approved):
@akhan-mozilla The schema and migrations are beautiful! 👩‍🍳🤌💋 Thanks for taking the time to iterate on this PR

I'm going to Approve as this looks pretty much ready to go. 

I did find a hopefully small bug when I tried to create a new Campaign from Django Admin using the plus button on the left pane as well as the one from the upper right when on the Campaigns list view. 

Screenshot 2024-09-16 at 6 10 14 PM

Hopefully it's a small one and you can move on to the React UI. It's definitely good to keep the Django Admin UI functional, but I'm eager to put most of our efforts into the React side.

PR #243, opened 2024-09-11
[AE-486] First-pass React development setup for AdOps Dashboard

mashalifshin:
## References

[AE-486](https://mozilla-hub.atlassian.net/browse/AE-486)

## Problem Statement

We need to set up a new React frontend in shepherd for the AdOps Dashboard, as well as a separate place for our API endpoints that will drive this UI. We want to be able to develop it alongside the existing shepherd templated frontend and admin functionality.

## Proposed Changes

This change adds a first-pass development setup for a React frontend and json REST API. 

Several decisions about tooling and repo organization are made in this PR. I'll call them out explicitly and I welcome any discussion/suggestions around these decisions.

- New directory for the new frontend, called `./dashboard`
- In `./dashboard`, Set up Vite + React + Typescript with SWC (speedier, newer than Babel)
- Set up `Dockerfile` and `docker-compose` with the new dashboard app so a single `docker-compose` command from the project root spins up everything. Changes developers make are hot-reloaded into the docker container
- Sets up a json REST API on the django end to serve data to the dashboard. The dashboard overviews aren't really RESTful resources, but I wanted to go with a very standard API choice with lots of tooling/support.
- Provisional App.jsx that demos that we can fetch from the API
- `react-query` to wrap up and consolidate common data-fetching concerns like cacheing, retries. This can also give some state management but I haven't gotten there yet.

## Verification Steps

Run `docker-compose up --build` from the root of the repo
Visit http://0.0.0.0:7001/api/v1/products to see Django's API debug pages
Visit http://0.0.0.0:5173/ and make sure you can see the React UI render a list of the BoostrProducts

## Next Steps

This is very bare bones currently so it can unlock others developing on the dashboard, and it still needs quite a bit of additional setup, which I'll add incrementally in subsequent PRs.

- jest test setup and first-pass unit test
- Add dashboard linting and tests to the existing Makefile
- Build for production and serve React app from Django

## Visuals

Screenshot 2024-09-11 at 4 07 20 PM
Screenshot 2024-09-11 at 4 09 44 PM


consvc_shepherd/api/views.py
```diff
@@ -0,0 +1,15 @@
+"""Dashboard API views that produce json data"""
+
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from consvc_shepherd.api.serializers import BoostrProductSerializer
+from consvc_shepherd.models import BoostrProduct
+
+
+@api_view(["GET"])
```
mashalifshin, line 10: We probably don't need this particular endpoint for the dashboard overviews/summaries, so this is just here temporarily to bootstrap a quick endpoint for the React frontend to hit and verify an e2e setup

consvc_shepherd/api/views.py
```diff
@@ -0,0 +1,15 @@
+"""Dashboard API views that produce json data"""
+
+from rest_framework.decorators import api_view
```
akhan-mozilla, line 3: This all looks good. Just an FYI, we can use `ModelViewSets` instead of `api_view` for the APIs, as they come with built-in CRUD functionality, eliminating the need to define each endpoint manually. This can be implemented when adding the APIs for the campaign summar

akhan-mozilla (approved):

consvc_shepherd/api/views.py
mashalifshin: Thanks for taking a look Ayesha! And for the helpful tip on `ModelViewSets`, I will definitely check that out and see if I can use that instead of the one-off test endpoint here

nanagb: I 2nd the suggestion for ModelViewSets I used that when I was experimenting with creating the API and it's a lot easier to work with.

PR #244, opened 2024-09-11
Shep 38 makefile improvements

nanagb:
## References

[SHEP-38](https://mozilla-hub.atlassian.net/browse/SHEP-38)

## Problem Statement

I've noticed a few tasks we  all have to do that are frequent or take multiple steps that are often forgotten when needed. I've added them to our Makefile to streamline development tasks.

I also cleaned up the formatting & arrangement  of the .PHONY commands and improved some of the help descriptions

## Proposed Changes
 modified `make makemigrations` behavior it will now make migrations and automatically migrate. If you just want to create migrations without the migrate run `make makemigrations MIGRATE=false`

added `make ruff` this uses the newer faster ruff linter instead of the multiple tools we currently use. The setting are still being tuned to match the older settings.

added `make debug` uses [docker debug](https://docs.docker.com/reference/cli/docker/debug/) to provide a more full feature cli for our container 

added `make remove-migration-fix VER=WXYZ` this needs a better name. But this simplifies the task of consolidating all the accumulated migration files into one file before merging. With one command you can revert migrations (back to VERY=WXYZ), delete the migration new unneeded files, create a new consolidated migration file and apply it.

## Verification Steps

Test these commands on your local dev env and send feedback on any modifications or update that may be needed.

[SHEP-38]: https://mozilla-hub.atlassian.net/browse/SHEP-38?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,117 +3,140 @@ COV_FAIL_UNDER := 95
 INSTALL_STAMP := .install.stamp
 POETRY := $(shell command -v poetry 2> /dev/null)
 VER :=
+MIGRATE ?= true
 
 # This will be run if no target is provided
 .DEFAULT_GOAL := help
 
-.PHONY: help
+.PHONY: help install isort isort-fix black black-fix flake8 bandit pydocstyle mypy lint lint-fix format local-migration-check local-migrate  test doc-install-deps doc doc-preview dev local-test makemigrations-empty migrate makemigrations remove-migration debug
+
 help: ##  show this help message
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m\033[0m\n"} /^[$$()% 0-9a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-16s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-.PHONY: install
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -3,117 +3,140 @@ COV_FAIL_UNDER := 95
...
-.PHONY: remove-migration
-remove-migration:  ##  Run command to undo migrations add VER= to set the migration number e.g. 0002
-	docker exec -it consvc-shepherd-app-1 python manage.py migrate consvc_shepherd ${VER}
\ No newline at end of file
+	@if [ "$(MIGRATE)" = "true"]; then \
+		echo "Applying migration..."; \
+		docker exec -it consvc-shepherd-app-1 python manage.py migrate; \
+	fi
+
+.PHONY: ruff
```
kauana, line 105: Perhaps we could also change ruff to adhere to the new format (remove `.PHONY` here and add it to line 11)?

Makefile
```diff
@@ -3,117 +3,140 @@ COV_FAIL_UNDER := 95
...
+ruff: install
```
kauana, line 106: It would be nice to add a help comment to it as well

Makefile
nanagb: Fixed thanks

Makefile
nanagb: Added a new help comment

kauana (approved):
LGTM!

PR #248, opened 2024-09-13
[SHEP-37] Fix the way the sync script updates existing Deals

mashalifshin:
## References

[SHEP-37](https://mozilla-hub.atlassian.net/browse/SHEP-37)

## Problem Statement

We're seeing [errors in deployed environments](https://mozilla.sentry.io/issues/5805480794/?project=4505206579462144&query=IntegrityError&referrer=issue-stream&statsPeriod=14d&stream_index=0) where the sync script succeeds in creating the initial Product entry the first time through, but fails to update existing Product on subsequent runs. That particular issue with Product upsert was fixed, but there is a similar problem with the way we're upserting Deals. 

## Proposed Changes

The usage of Django's `update_or_create` was incorrect for upserting Deals, and this PR updates to the correct usage.
I also separated out the mocks into their own files outside the test file, to make it a bit more manageable to work with the tests.

## Verification Steps

These specs verify the new usage, but they don't explicitly verify that when the script encounters an existing deal, the upsert behavior is correct. To do this, I'd want to refactor the `upsert_deals` function into several pieces so it'd be easier to test this kind of behavior, but I held off on that in this PR. 

[SHEP-37]: https://mozilla-hub.atlassian.net/browse/SHEP-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nanagb (approved):

mashalifshin:
@nanagb thanks for the review! In the future, could you please make a comment or use github's "suggested changes" feature to request changes on my branch? I'm used to having a discussion before other contributors push to my branch so I'm not surprised. But maybe that's just what I'm used to, we can discuss our git PR and review process in our next team norming.

I reverted the commit with the test fix that you added to this branch, because it's not the way I want to approach the fix (and because it's out of scope for this change, I'd like to tackle in a separate PR). 

The skipped test now passes because we added the max deal pages limit, so with your commit, the test behavior is that the deals loop runs 50times, up to the limit, and then finishes, passing the test, but not for the right reason (and also making the test take longer than it needs to). 

We already have other tests that assert that the max deal pages limit is working, so what I really want this test to verify is the behavior that when the script receives an empty array response from the Boostr API, the loop finishes correctly. So we want to make our mock a little smarter to support that, or use a different mock that just returns empty response, and then the test will pass for the right reason.

I made a ticket so we don't forget about this https://mozilla-hub.atlassian.net/browse/SHEP-41, one of us can come back to it when we need a quick win 😁

PR #249, opened 2024-09-13
[SHEP-27] Rate limit adjusted GET & POST using retry-after header info

nanagb:
## References

[SHEP-27](https://mozilla-hub.atlassian.net/browse/SHEP-27)

## Problem Statement

Boostr's API has a 100 reqs per min rate limit that's inadequate for our syncing needs. During full syncs, we experience rate limits (429 errors). Our initial solution to introduce rests between each call  didn't always work. Likely due to how rate limits are handled by the Boostr API. 

## Proposed Changes

The 429 response includes a Retry-After header with the number of seconds to wait for before retrying. The new POST and GET functions uses this to determine how long to wait before retrying the sync. An extra second is added to the Retry-After time as an extra precaution. If no value is readable from the header, it will retry again in 60 seconds, which is enough time for the rate limit counter to fully reset.

## Verification Steps

Clear your DB tables and do a full sync (Approx run time should be 3.5 to 4 mins). If successful, there will be no 429 errors in the logs. 

[SHEP-27]: https://mozilla-hub.atlassian.net/browse/SHEP-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +112,51 @@ def authenticate(self, email: str, password: str) -> str:
...
+        if headers is None:
+            headers = {}
+
         response = self.session.post(
             f"{self.base_url}/{path}",
             json=json,
             headers=headers,
             timeout=15,
         )
+
```
mashalifshin, line 130: It looks like we are no longer sleeping between requests at all, which is great because the script goes so much faster! 

But as part of this PR, we also need to remove everything related to the request rate interval (the default constant, the config option, the parameter to the script, and the instance variable in the `BoostrApi` class, and all the documentation about it). We don't want to leave around an option that does nothing.

I think we'll also want a PR against [the cron job in the cloud infra repo](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/topsites/k8s/charts/shepherd/values.yaml) to remove that option from the invocation of the script, and maybe some updates to values?. We may want to merge that change before we merge this PR where we remove it entirely (if we think the cron will error when passed a non-existent option, we can check on this locally).

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +112,51 @@ def authenticate(self, email: str, password: str) -> str:
         token = self.post("user_token", post_data)
         return str(token["jwt"])
 
-    def post(self, path: str, json={}, headers={}) -> Any:
+    def post(self, path: str, json=None, headers=None) -> Any:
```
mashalifshin, line 115: Still a python newbie, can you help me understand the change here and related change below to set the json param to `{}` when it's `None`? I thought that specifying the parameter with the equals sign like `json={}` makes `{}` the default value for `json` if nothing is passed in? Am I missing something?

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +112,51 @@ def authenticate(self, email: str, password: str) -> str:
...
-        sleep(self.request_interval_seconds)
+        if params is None:
+            params = {}
+        if headers is None:
+            headers = {}
+
         response = self.session.get(
             f"{self.base_url}/{path}", params=params, headers=headers, timeout=15
         )
+        if response.status_code == 429:
```
mashalifshin, line 155: Could we add a debug log here that we received the 429, and what the wait time is?

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +112,51 @@ def authenticate(self, email: str, password: str) -> str:
...
+            retry_after = int(response.headers.get("Retry-After", 60)) + 1
+            time.sleep(retry_after)
+            return self.get(path, params, headers)
```
mashalifshin, line 158: I think I'm confused by the control flow here? Is there a reason we're returning this instead of doing something like 

```python
response  = self.get(path, params, headers)
```

Also you can use `self.session.get` here, which automatically holds the auth headers. I don't think it actually matters since we're also passing the auth headers explicitly, but it's just nice to be consistent

consvc_shepherd/tests/test_sync_boostr.py
```diff
@@ -382,9 +382,8 @@ def mock_update_or_create_deal_product(
 class TestSyncBoostrData(TestCase):
```
mashalifshin, line 382: I think we need to add a new test here that verifies the correct handling of 429 responses. Lmk if you want to pair or if I can support you in creating that in some other way. We'd want a new kind of MockResponse that returns a 429, and then we can assert that the sleep function was called with the right amount of time.

mashalifshin (changes requested):
Thanks so much for peeling this off into its own PR, super helpful! And for making the fix, I ran it locally and this approach is working much better than trying to wait between requests!

To make this merge-able there is a bit more work to do on adding specs and cleaning up the request rate interval option that got disabled (both here and in the cloudops infra repo). Also needs a new test verifying the 429 handling behavior. Details in the comments.

mashalifshin:
Oh yes and don't forget to add [SHEP-27] to the title of the PR (check out `git log --oneline` to see how nice this comes out in the repo history).

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: I also recently discovered this. But basically default params are evaluated once when function definitions are evaluated and not every time the function is called. So when you have a mutable such as a dict or list it can cause issues because it holds onto old values during subsequent calls.

There are two helpful posts here:
[https://web.archive.org/web/20200221224620id_/http://effbot.org/zone/default-values.htm](https://web.archive.org/web/20200221224620id_/http://effbot.org/zone/default-values.htm)
[https://docs.python.org/3/reference/compound_stmts.html#function-definitions ](https://docs.python.org/3/reference/compound_stmts.html#function-definitions ) Read the section titled **"Default parameter values are evaluated from left to right when the function definition is executed."**

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: Done

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: It's a recursive call to make sure all the logic around 429 errors are applied with each attempt. `self.session.get` would not handle a 429 that gets thrown right after sleeping. This could happen if some other process triggered a new 429 during the sleep.

nanagb: `response  = self.get(path, params, headers)` also works I just used return to skip the extra step of returning the response

consvc_shepherd/tests/test_sync_boostr.py
nanagb: Taking a first pass at it, I'll ping you if I hit a roadblock.

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: Done. I removed it right before my commit 😃

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +100,70 @@ def authenticate(self, email: str, password: str) -> str:
...
+            headers = {}
+
         response = self.session.post(
             f"{self.base_url}/{path}",
             json=json,
             headers=headers,
             timeout=15,
         )
+
+        if response.status_code == 429:
```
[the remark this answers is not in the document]
nanagb: Sounds like a great idea, I'll make that mod

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +100,70 @@ def authenticate(self, email: str, password: str) -> str:
         token = self.post("user_token", post_data)
         return str(token["jwt"])
 
-    def post(self, path: str, json={}, headers={}) -> Any:
+    def post(self, path: str, json=None, headers=None) -> Any:
         """Make POST requests to Boostr that uses the session, pass through headers and json data,
         check status, and return parsed json
         """
-        sleep(self.request_interval_seconds)
+        if json is None:
```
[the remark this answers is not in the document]
nanagb: Good call, editing

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +100,70 @@ def authenticate(self, email: str, password: str) -> str:
...
-            f"{self.base_url}/{path}", params=params, headers=headers, timeout=15
-        )
+        if params is None:
+            params = {}
+        if headers is None:
+            headers = {}
+
+        try:
+            response = self.session.get(
+                f"{self.base_url}/{path}", params=params, headers=headers, timeout=15
```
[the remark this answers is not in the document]
nanagb: Done ty

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: @cyptm No I didn't see it, but based on the documented behavior and our code, we likely would have been affected. As mentioned, I stumbled upon this, I think it popped up while testing out the ruff linter, so I went searching for why this was such a big deal.

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +102,67 @@ def authenticate(self, email: str, password: str) -> str:
...
+                        Current retry: {current_retry}"
+                    )
+                    time.sleep(retry_after)
+                    current_retry += 1
+                    return self.get(path, params, headers, max_retry, current_retry)
+                else:
+                    raise BoostrApiError("Maximum Retries Reached")
+        except requests.exceptions.RequestException as e:
+            if current_retry < max_retry:
+                self.log.info(f"{e}: Read timeout Current Retry: {current_retry}")
```
mashalifshin, line 160: Let's remove the "read timeout" part of this log, because it could be any sort of exception thrown from requests that is being handled here, not just the read timeout, right?

mashalifshin (commented):
Nana it works really well!!!!!! Was able to run a full sync successfully in ~23 minutes.  🙌

the-rock-snl-hahaha-it-works

Though I did not encounter any connection closed or timeout issues this run, just the rate limits response, so not sure how it would work with that.

There are a lot of comments here but the main points are that to be ready for merge, this PR needs some better testing of the various possibilities/branches in the retry code, and ideally we would clean up the retry control flow since it's hard to reason about and also hard to test.

I know I wrote a lot of text, but I'm just trying to be as helpful as I can. Lmk if pairing or talking it over would work better for you.

Makefile
```diff
@@ -103,8 +103,11 @@ makemigrations: ##  Run makemigrations on the docker container set MIGRATE=false
 	fi
 
 ruff: install ##  **Experimental** Run ruff linter. To fix and format files.
+	@echo "Running Ruff..."
 	$(POETRY) run ruff check --select I --fix
 	$(POETRY) run ruff format
+	@echo "Running Mypy..."
+	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
```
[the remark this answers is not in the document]
nanagb: I agree with your point. We can do it either way, I think it makes sense to move it to a separate PR.

One reason it's here is because `make ruff` is not officially part of any of the automated linting we are doing in CI/CD.  `mypy` is already part of the commands we have in `make lint` and `make lint-fix` so I added it because it was missing.

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: @mashalifshin Based on my understand of how this works, this is the line that could have been a potential cause of bugs https://github.com/mozilla-services/consvc-shepherd/blob/df3f3a2906183a81f12f9f1e0ca3890b4b48c09d/consvc_shepherd/management/commands/sync_boostr_data.py#L164

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +102,67 @@ def authenticate(self, email: str, password: str) -> str:
...
+                    self.log.info(
+                        f"{response.status_code}: Rate Limited - Waiting {retry_after} seconds. \
+                        Current retry: {current_retry}"
+                    )
+                    time.sleep(retry_after)
+                    current_retry += 1
+                    return self.get(path, params, headers, max_retry, current_retry)
+                else:
+                    raise BoostrApiError("Maximum Retries Reached")
+        except requests.exceptions.RequestException as e:
```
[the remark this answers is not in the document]
nanagb: Yes correct for both comments. I thought about having rate limits and the timeouts having their own retries, but in the worst case that could end up with a get request that take more than 5 mins to complete. If somehow multiple requests end up in this situation, I can see it taking multiple hours or days to run a sync.

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -112,32 +102,67 @@ def authenticate(self, email: str, password: str) -> str:
...
+                headers=headers or {},
+                timeout=15,
+            )
+            if response.status_code == HTTP_TOO_MANY_REQUESTS:
+                if current_retry < max_retry:
+                    retry_after = (
+                        int(response.headers.get("Retry-After", MAX_SLEEP_SECONDS)) + 1
+                    )
+                    self.log.info(
+                        f"{response.status_code}: Rate Limited - Waiting {retry_after} seconds. \
```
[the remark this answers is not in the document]
nanagb: Yep not the intended result. I was formatting the line for the linter because it was over 120 chars. Fixed now

consvc_shepherd/management/commands/sync_boostr_data.py
nanagb: Yes, this is a generic exception for multiple types of errors. Fixed

consvc_shepherd/tests/test_sync_boostr.py
```diff
@@ -42,28 +43,36 @@ def test_setup_session(self, mock_sleep, mock_authenticate):
...
         jwt = boostr.authenticate(EMAIL, PASSWORD)
         self.assertEqual(jwt, "i.am.jwt")
 
-    @mock.patch("consvc_shepherd.management.commands.sync_boostr_data.sleep")
+    @mock.patch(
+        "consvc_shepherd.management.commands.sync_boostr_data.BoostrApi.get",
+        side_effect=mock_get_error_429,
+    )
+    @mock.patch("requests.Session.post", side_effect=mock_post_success)
+    def test429(self, mock_post, mock_get):
```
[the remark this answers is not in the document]
nanagb: I'll make some more mods to this test and check in with you tomorrow to setup a time to pair and review. Thanks

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -48,14 +46,6 @@ def add_arguments(self, parser):
                 {MAX_DEAL_PAGES_DEFAULT} pages. Currently we have ~14 pages of deals in Boostr, so this default max
                 should be sufficient for a while.""",
         )
-        parser.add_argument(
-            "--request-interval-seconds",
-            default=REQUEST_INTERVAL_SECONDS_DEFAULT,
```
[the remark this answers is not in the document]
mashalifshin: Good find, thank you!

mashalifshin (approved):
🚢 Well tested!!! Thank you

PR #253, opened 2024-09-19
SHEP18: Add DeliveredFlight model & migration and display DeliveredFlight data on CampaignSummary view

kauana:
## References

[SHEP-18](https://mozilla-hub.atlassian.net/browse/SHEP-18)

## Problem Statement

We need to create a new model and migration for the kevel data we will pull from BigQuery. Our `CampaignSummary` view will need to display click/impression information including clicks/impressions delivered, CTR (Clicks Delivered / Impressions Delivered) and Live (Impressions_Delivered > 0).

## Proposed Changes

1. **Add a new script `seed.py`**: this script generates random data to populate our models, which is then displayed on the admin page. By running `python manage.py seed` (within Shepherd's container), you can view the mock data on the admin dashboard. Currently, the script generates data for all existing models, but any future models will need to be added to `seed.py`. This script is optional but has been very helpful for testing models, so I believe it could benefit the team. Check the visuals for an example of what it generates.

1. **Add the DeliveredFlight model**: This model will contain the flight related data ingested from BigQuery (the script to pull from BigQuery will come in a later PR).

1. **Changes to `admin.py`**: There is a new admin model for delivered flights. It is also  possible to filter the delivered flights by partner name, which will be useful once we start parsing data from other partners such as adm, and filter by date. 

    The `Campaign Summaries` admin model also displays the sum of clicks/impressions per flight ID, whether campaign is live and CTR.

1. **Add Unit tests**: Add model and admin tests  for `DeliveredFlight`. The admin tests also include tests for the 2 filters included: Partner filtering and submission date filtering.

8. **Documentation for seed.py**: Add documentation for running the seed custom command in our docs.

## Verification Steps

1. As always, run our tests (`make local-test`) and ensure they pass.
2. Run the migrations to create the new database table
3. Use the seed command to generate some mock data, then take a look the `Delivered campaigns` and the `Campaign Summaries`. 

## Visuals
DeliveredCampaign (populated with seed script):
image

Partner filter for delivered campaign:
image

Campaign Summaries with Impression/Click data, etc.:
image

[SHEP-18]: https://mozilla-hub.atlassian.net/browse/SHEP-18?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/models.py
```diff
@@ -432,24 +432,104 @@ def __str__(self):
...
+            return self.impressions_sold - self.impressions_delivered
+        return 0
+
+    @property
+    def live(self):
+        """Whether the campaign is active"""
+        if self.impressions_delivered > 0:
+            return "Yes"
+        return "No"
+
```
akhan-mozilla, line 484: Could you please update the test cases for the Campaigns model to include the new fields you added with the property decorator, similar to how the `net_ecmp` field is [handled](https://github.com/mozilla-services/consvc-shepherd/blob/d45e089ee63176ae9a0b1fd651697f2bcc7451c3/consvc_shepherd/tests/test_models.py#L117)?

consvc_shepherd/models.py
```diff
@@ -432,24 +432,104 @@ def __str__(self):
 
 
 class CampaignSummary(models.Model):
-    """Model representing a summary of campaign metrics."""
+    """Model representing a summary of campaign metrics including Boostr, and kevel
```
akhan-mozilla, line 435: Please update the docstring to include the new fields clicks_delivered and impressions_delivered you added.

consvc_shepherd/models.py
```diff
@@ -404,7 +404,7 @@ class Campaign(models.Model):
 
     ad_ops_person: CharField = models.CharField()
     notes: CharField = models.CharField()
-    kevel_flight_id: IntegerField = models.IntegerField()
+    kevel_flight_id: IntegerField = models.IntegerField(unique=True)
```
akhan-mozilla, line 407: We are currently using the `kevel_flight_id` as the identifier for campaigns. Since we will have multiple flights per campaign, we will need to update the existing schema accordingly and need to determine a identifier for campaigns. This can be addressed as part of the multiple flights functionality.

kauana: Exactly, this identifier (which I believe we will create) will also be useful when we start capturing metrics for other partners that don't have a flight_id (e.g. adm)

consvc_shepherd/models.py
kauana: Good catch, thank you!

consvc_shepherd/management/commands/seed.py
```diff
@@ -0,0 +1,153 @@
+"""This script generates mock data to test our models which get displayed in the Admin page"""
```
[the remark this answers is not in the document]
mashalifshin: Oh, would you mind please adding the info about seeding to the documentation, probably in the `quickStart.md` would be a great spot.

consvc_shepherd/models.py
```diff
@@ -432,24 +432,104 @@ def __str__(self):
...
+    Methods
+    -------
+    __str__(self)
+        Return the string representation for a Delivered Campaign
+
+    """
+
+    submission_date: DateField = models.DateField()
+    campaign_id: IntegerField = models.IntegerField()
+    flight: ForeignKey = models.ForeignKey(
```
mashalifshin, line 523: I think I might be getting confused by the naming here, or maybe the doc string comment above, but I was thinking that we'd store `flight_id` as the char field here, and that would be the `flight_id` that is coming from BQ, like all the data in this table. 

I was also imaging we would _not_ have any foreign keys from this table to the `Campaigns` table, and there would be no relational connection between `DeliveredCampaigns` and `Campaigns`.

And we would put the data together in the `CampaignSummary` DB view. We'd pull the `DeliveredCampaigns` by the flight id (coming from BQ), as well as pull in the `Campaigns` data, and associated Deal and Product data, by flight id (manually entered by the AdOps people)

I do want to understand if there is utility in having a foreign key into the `Campaigns` table here. Though we'd want to associate the data here with a particular flight, not a particular campaign, in light of what we learned from Laura re. wanting to be able to have multiple flights in a Campaign.

Do we want to have another schema chat together with @akhan-mozilla? I will also be working on diagramming our schema as part of writing up that tech spec, which will help us game out some possible options as well

mashalifshin (commented):
@kauana Awesome work here! Loving the new seeder but I am getting an error when trying to run it. I see it's added as a dev dependency so I might just not know how to do that since I'm new to python ecosystem? Or maybe there is something to debug about how this dependency is pulled into the docker container for dev? That part I can help with more better ;-)  


Here's what I did
1. Built your branch:  `docker-compose up --build`
1. Shell into running container:  ` docker exec -it consvc-shepherd-app-1 sh`
1. From the container's shell, run: ` python manage.py seed`

Getting

```
  File "/app/consvc_shepherd/management/commands/seed.py", line 8, in <module>
    from faker import Faker
ModuleNotFoundError: No module named 'faker'
```

And I want to talk a bit more about the schema, left a comment on there but maybe we need to meet up again to really make sure we've got it right. Especially in light of new info that a Campaign could contain multiple flights ...

consvc_shepherd/models.py
akhan-mozilla: You’re right that we can retrieve the `campaignsDelivered` and `campaigns` data together in the campaign summary without a direct relationship, since we don't have a single source for the flight IDs for both tables. As you mentioned, for `campaignsDelivered` it will come from BigQuery (BQ), and for `campaigns`, Ad Ops will need to enter it manually. However, it would be helpful to have a diagram and clearly defined foundational rules before proceeding with the implementation.

consvc_shepherd/management/commands/seed.py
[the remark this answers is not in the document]
kauana: Hey Masha, thank you for trying it out. 

I was including faker in the wrong section in `pyproject.toml` 🤦‍♀️ . When you get the chance, could you please pull the latest commit, try again and let me know if it works?

consvc_shepherd/models.py
kauana: Hey Masha, 

Yeah, I think the doc string could be improved. The code I have does exactly what you described. Delivered campaigns will have a `flight_id`. You can actually see how I pull the `flight_id` from BQ on the script I have in progress here. In fact, if you take a look at the `consvc_shepherd_delivered_campaigns`table on pslq, you will see that we have a flight_id field that is of `IntegerType`:

image

I decided to create a foreign key in `DeliveredCampaigns` to the `kevel_flight_id` in `Campaigns` to increase the relational integrity between the two tables and avoid orphaned records in the `DeliveredCampaigns` table in case a Campaign is deleted or doesn't exist to begin with. In additon, by having a foreign key here we will automatically have an index generated for us and the join in the `CampaignSummary` view will be more performant.

Having said that, I'm down to change the `flight_id` into an `IntegerField` and remove the foreign key as well. 


> Do we want to have another schema chat together with @akhan-mozilla? I will also be working on diagramming our schema as part of writing up that tech spec, which will help us game out some possible options as well

This is a good question. I would be down to meet up and discuss how we will split these campaigns into multiple flights.

consvc_shepherd/models.py
kauana: Hi @akhan-mozilla, I have added the tests for the `CampaignSummary` view to test the calculated fields. Note I've changed `net_epcm` slightly to round the results to 2 decimals to avoid long floats. Let me know what you think, thanks.

consvc_shepherd/models.py
mashalifshin: I think the identifier for Campaigns can be the uuid that's local to shepherd here, since Campaigns is a local-to-shepherd concept we're creating.

Looking ahead to non-kevel providers, I think we could do a "polymorphic type" sort of thing, which would be to rename `kevel_flight_id` to a more general `flight_id`, and then add another column, `flight_partner`, which would specify `kevel` or `adm` or whatever.

consvc_shepherd/models.py
mashalifshin: I made some time for us to connect tomorrow and talk through all of it. I'll create some diagrams ahead of time so we can see the options clearly and think through some tradeoffs

consvc_shepherd/admin.py
```diff
@@ -393,6 +404,53 @@ class CampaignSummaryAdmin(admin.ModelAdmin):
...
+    """Filter for listing by partner."""
+
+    title = _("partner")
+    parameter_name = "partner"
+
+    def lookups(self, request, model_admin):
+        """Return a list of distinct partners for the filter options."""
+        # Define the possible values for partner
+        return [
+            ("Contile", "Contile"),
```
mashalifshin, line 417: Oh yes I wanted to mention that [Contile](https://mozilla-hub.atlassian.net/wiki/spaces/MA1/pages/449183855/Contile) isn't a 3rd party partner, it's a deprecated internal service that used to talk to ADM for getting Tiles and has been replaced by MARS, so maybe we should use ADM here or something. 

(There are still references to Contile all over the place, I want to double check with ads-eng team but I think we can start cleaning those up)

docs/operations/seed.md
```diff
@@ -0,0 +1,42 @@
...
+
+As new models are added in the future, we should consider including them in the seed script to make it easier to see how all the models interact.
+
+## How to Run
+
+1. Run Shepherd locally and exec into the Shepherd container to run the script:
+    ```shell
+    docker exec -it consvc-shepherd-app-1 bash # can also run with `make debug`
+
+    docker manage.py seed
```
mashalifshin, line 15: Such thorough documentation, thank you! 📜 

Just found a small but key type here: `python manage.py seed`

docs/quickStart.md
```diff
@@ -4,93 +4,92 @@ To use consvc-shepherd, you'll need a Python 3.11 development environment with P
...
 While the Django Admin interface at /admin shows all the data in our DB in a human readable way, it may sometimes be
 helpful to connect directly to the Shepherd DB while developing. (The values for the DB user and DB name come from
 the `.env` file.)
 
-```sh
-docker exec -it consvc-shepherd-db-1 psql -U postgres postgres
-```
+    ```shell
+    docker exec -it consvc-shepherd-db-1 psql -U postgres postgres
+    ```
```
mashalifshin, line 78: Thanks for cleaning up this markdown! Maybe let's mention the seeder here, before we mention syncing boostr data, since that's a much quick-erStart for boostrapping local dev

consvc_shepherd/management/commands/seed.py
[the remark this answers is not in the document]
mashalifshin: So on the latest commit, it gets further and is able to find the `faker` dep, but I'm getting a stack trace. I suspect the difference is that you started with the empty DB, and I'm starting with a DB full of data from the Boostr sync ...

```
Traceback (most recent call last):
  File "/app/manage.py", line 22, in <module>
    main()
  File "/app/manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/consvc_shepherd/management/commands/seed.py", line 81, in handle
    BoostrDealProduct.objects.get_or_create(
  File "/usr/local/lib/python3.11/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/models/query.py", line 916, in get_or_create
    return self.get(**kwargs), False
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/models/query.py", line 640, in get
    raise self.model.MultipleObjectsReturned(
consvc_shepherd.models.BoostrDealProduct.MultipleObjectsReturned: get() returned more than one BoostrDealProduct -- it returned 3!
```

Also lol at the exclamation point here, I like how excited the stack trace sounds about finding 3 `BoostrDealProducts` 😂

consvc_shepherd/admin.py
kauana: Yup, I've heard it's deprecated, but the reason I used Contile here is because that's what Glean is still reporting. That's a good point though because since we are focusing on Kevel data first, we shouldn't have any `contile` metrics showing up. I will go ahead and change it to ADM instead

docs/operations/seed.md
kauana: Good eyes!

consvc_shepherd/management/commands/seed.py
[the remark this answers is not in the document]
kauana: Hmm you are correct. To be honest, I've only run the  boostr sync script once. Is this something we need to run frequently?

docs/quickStart.md
kauana: Done!

consvc_shepherd/management/commands/seed.py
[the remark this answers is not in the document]
mashalifshin: I don't think you really need to run it frequently unless you're developing on it, maybe just once to populate some real data into your development DB. Though an easier and faster way to do that would be to grab a `pgdump` from staging or prod.

But in terms of this particular PR, I think the seed script should be agnostic to pre-existing data in someone's local DB and still work if they have run the sync script, right? Or were you thinking this is just for folks who don't have any local data?

docs/quickStart.md
mashalifshin: Also I super appreciate the thoughtful clarifications you've added throughout this file!

mashalifshin (approved):
:shipit: Looking good! I just spot checked the seed script and it's working great now even with all that boostr data in my local DB. Also checked that things look good on the Django admin side, and the model and DB is looking good as well.

kauana:
@mashalifshin thanks for checking the `seed` script again! Removing the foreign key from `DeliveredFlight` fixed it because now we don't depend on Campaigns being created for BoostrDeals 🎉

PR #259, opened 2024-09-24
fix(GHA workflows): set correct step id in docker-login action

ahoneiser:


ahoneiser:
:information_source: also updating tag pattern with this PR, based on discussions in https://github.com/mozilla-services/mars/pull/250

mashalifshin (approved):

PR #263, opened 2024-09-25
[SHEP-56] Set up jest testing and typescript eslint for the ad-ops-dashboard

mashalifshin:
## References

[SHEP-56](https://mozilla-hub.atlassian.net/browse/SHEP-56)

## Problem Statement

We need to set up linting and testing for the new React dashboard. 

## Proposed Changes

In this change, we
- Update the `Makefile` to include testing and linting the dashboard when running the usual `make local-test`, `make test` and `make lint` commands
- Use python-node images in CircleCI so that we can get node/npm out of the box for those builds
- Configure eslint for typescript and vite
- Configure jest testing and dashboard code coverage (though leaving requirement of coverage commented out for now)
- Add a first-pass jest test
- Clean up the products-fetching code 

## Verification Steps

Please run: 
- `make local-test` locally and verify both the shepherd and dashboard test suites are running fully and correctly
- `make eslint` and `make eslint-fix` locally and verify that dashboard is being linted properly. Left the `eslint` command out of the `make lint` command for now since we have some additional lint fixes that need to go in before this passes, and this PR is large enough
- Check the CI output for the latest, successful runs for the `lint` and `test` steps and verify the same as above

[SHEP-56]: https://mozilla-hub.atlassian.net/browse/SHEP-56?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dashboard/src/__ tests __/App.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+describe("<App />", () => {
+  test("Sanity check", () => {
+    expect(true).toBe(true);
+  });
+
+  test("shows the loading screen", () => {
+    render(<App />);
+    const header = screen.getByText("Loading...");
+    expect(header).toBeInTheDocument();
```
mashalifshin, line 13: This test needs to mock out `fetch` to do anything more fun, but I punted on that in the interest of getting this initial setup out sooner. You'll see that the component logs an error, but the test still passes, since this aspect is unfinished ...

docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
     tty: true
     env_file:
       - .env
+    image: app
```
mashalifshin, line 18: I believe adding this gives the built image a name, which means we rebuild the named image each time instead of leaving around lots of anonymous dangling images and filling up docker storage and getting `Error response from daemon: max depth exceeded` when trying to build.

Before:

Screenshot 2024-09-25 at 4 13 58 PM

After:

Screenshot 2024-09-25 at 4 16 59 PM

kauana: Yup adding an image name here will help as docker will just overwrite the previous image. I don't think it cleans up failed builds though. We will still need to prune images from time to time (unless we write perfect code all the time that never fails :D) but that's probably not a huge deal.

Makefile
```diff
@@ -13,21 +14,24 @@ MIGRATE ?= true
 help: ##  show this help message
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m\033[0m\n"} /^[$$()% 0-9a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-16s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-install: $(INSTALL_STAMP)  ##  Install dependencies with poetry
+install: $(INSTALL_STAMP)  ##  Install dependencies with poetry and npm
 $(INSTALL_STAMP): pyproject.toml poetry.lock
```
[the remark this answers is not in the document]
mashalifshin: I played around with this a bit and ended up going with installing both under the same command, and sharing the install stamp. This way they share the install stamp, which I think is cleaner/more convenient here than having two different install stamps. I guess the logic here is that they are still one app with a python part and a node part, and we'll always lint, test, and install them together.

Lmk if I'm answering your question, I wasn't sure if I understood it properly since I'm not sure which block of code you meant exactly

docker-compose.yml
mashalifshin: > (unless we write perfect code all the time that never fails :D)

BWAHAHAHAHA. But yes great point, we'll still get those un-named images for failed builds. But yeah I think it's still gonna be helpful so we don't need to prune as often, and it sounds like you agree, so still a good change to make. 

> And just to confirm this would be using the same image and Dockerfile as the main app correct?
Yeah setting this doesn't change how the image is build except for giving it a name. The build clause below still specifies that we'll use the `Dockerfile` at the root to build this. And we haven't touched the actual `Dockerfile`, just the `docker-compose.yml`, so this change only applies in development

kauana:
Few observations:

1. We should add steps to our docs/quickstart to download `node` as well now
2. Running `make local-test` doesn't work for me. I get the error below:
    ```
    console.error
    dashboard-1  |     ReferenceError: fetch is not defined
    dashboard-1  |         at getProducts (/dashboard/src/queries.tsx:5:22)
    ...
    
    dashboard-1  |        7 |     return products;
    dashboard-1  |        8 |   } catch (error) {
    dashboard-1  |     >  9 |     console.error(error); 
    dashboard-1  |          |             ^
    dashboard-1  |       10 |     return [];
    dashboard-1  |       11 |   }
    dashboard-1  |       12 | };
    ```
    
  3. `make eslint / make eslint-fix` works for me

Makefile
```diff
@@ -84,16 +99,20 @@ doc-preview: doc  ##  Preview Merino docs via the default browser
 dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
 	docker compose up
 
-local-test: $(INSTALL_STAMP)  ##  local test
-	docker compose -f docker-compose.test.yml up --abort-on-container-exit
+local-test-shepherd: $(INSTALL_STAMP)
+	docker compose -f docker-compose.test-shepherd.yml up --abort-on-container-exit
+
+local-test-dashboard:
```
kauana, line 105: we should add makefile help comments to all these new targets

kauana, line 105: nitpick: I think the `shepherd` and `dashboard` container names might be a bit confusing, as dashboard doesn't really suggest a connection to the front-end UI in my mind. Do you think we could rename this something like `frontend` `backend` or `django` 'js' or something like that?

mashalifshin: Good call, thank you! I think I've got `em all now, but let me know if you see any more I missed

mashalifshin: Yes! I played around with a couple options before reluctantly settling on these, I agree they could be better. 

I went with these here because they mirror the top-level directory structure, where we have a couple different apps defined -- though not quite since it's `consvc_shepherd` there. But I think of that as its full name, and `shepherd` as the first name (and shep as the nickname 😁)

Some other thoughts, lmk what you think is best:
- `frontend` and `backend` are ambiguous since the existing shepherd django app has a frontend already
- Could do `test-django` and `test-react`? Or `test-ts` and `test-python`? Both of those are named for the implementation details instead of what thing it's testing. And `test-python` isn't quite right here since we do test that django FE as well ...

More generally maybe the `dashboard` top-level name needs to be revisited to help clarify this? Like `ad-ops-dashboard` might be better?

Lmk what you think.

kauana: `test-django` and `test-react` sound good, and they clearly differentiate the two apps in my opinion.

> More generally, we might want to revisit the top-level name for the dashboard to make things clearer. For example, ad-ops-dashboard could be a better fit to distinguish it.

That’s a good point, and I agree it’s important to find a word to differentiate the two systems (react app vs django app) now, while we’re still early in the process. I’ve been referring to the django as the "Shepherd UI," but I realize that’s going to be ambiguous moving forward.

I recall you mentioned that the Ad Ops team won’t be interacting with the Django app, so calling the react app `ad-ops-dashboard` makes it clearer to me.

kauana: Thanks for adding those comments, they look good.

mashalifshin:
> 1. We should add steps to our docs/quickstart to download `node` as well now

Yeah, great point. I added some instructions to `quickStart`, would you mind test-driving them on your end? Thank you

Makefile
```diff
@@ -40,38 +44,49 @@ flake8: $(INSTALL_STAMP)  ##  Run flake8
...
+	cd dashboard && npm run lint
 
-lint-fix: $(INSTALL_STAMP) isort-fix black-fix flake8 bandit pydocstyle mypy ##  Run various linters and fix errors to pass CircleCi checks
+eslint-fix: $(INSTALL_STAMP)  ##  Format code with eslint
+	@echo "Running eslint with autofix..."
+	cd dashboard && npm run lint:fix
+
+lint: $(INSTALL_STAMP) isort black flake8 bandit pydocstyle mypy eslint ##  Run various linters
+
+lint-fix: $(INSTALL_STAMP) isort-fix black-fix eslint-fix flake8 bandit pydocstyle mypy eslint-fix ##  Run various linters and fix errors to pass CircleCi checks
```
nanagb, line 69: Did you intentionally do eslint-fix twice? If yes, can you explain what's happening? thanks

Makefile
```diff
@@ -13,21 +14,24 @@ MIGRATE ?= true
...
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m\033[0m\n"} /^[$$()% 0-9a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-16s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-install: $(INSTALL_STAMP)  ##  Install dependencies with poetry
+install: $(INSTALL_STAMP)  ##  Install dependencies with poetry and npm
 $(INSTALL_STAMP): pyproject.toml poetry.lock
+	# Python dependencies
 	@if [ -z $(POETRY) ]; then echo "Poetry could not be found. See https://python-poetry.org/docs/"; exit 2; fi
 	$(POETRY) install
+	# Node dependencies
+	@if [ -z $(NPM) ]; then echo "Npm could not be found. See npm install instructions for you platform"; exit 2; fi
```
nanagb, line 23: Nit: typo "your"

mashalifshin: Thank you!

Makefile
mashalifshin: Oh dear, no that was not intentional. Good eye, thank you!

Makefile
mashalifshin: Thanks for helping me out with the naming! I went with `test-django` and `test-react`, and then renamed `dashboard` -> `ad-ops-dashboard`

kauana (approved):
LGTM 🎉

PR #265, opened 2024-09-27
Shep 76 makefile improvements enhance ruff

nanagb:
## References

[SHEP-76](https://mozilla-hub.atlassian.net/browse/SHEP-76)

## Problem Statement

We currently use 6 different linting tools which can tend to be slow to run multiple times during the dev cycle.

## Proposed Changes

Experiment with consolidating most of the linters to [ruff](https://astral.sh/ruff) a modern fast linter and code formatter written in rust.  We'll benefit from a faster linting process and consolidation to just 2 tools (ruff and mypy)

## Verification Steps

Test `make lint-fix` and compare with `make ruff`
make ruff should fix everything make lint-fix can do, with a noticeable difference that it will format all line lengths to max 120 chars. See #216 for past discussion on this change

[SHEP-76]: https://mozilla-hub.atlassian.net/browse/SHEP-76?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,8 +103,11 @@ makemigrations: ##  Run makemigrations on the docker container set MIGRATE=false
 	fi
 
 ruff: install ##  **Experimental** Run ruff linter. To fix and format files.
+	@echo "Running Ruff..."
 	$(POETRY) run ruff check --select I --fix
 	$(POETRY) run ruff format
+	@echo "Running Mypy..."
+	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
 
 debug: ##  Connect to the shepherd container with docker debug.
 	docker debug consvc-shepherd-app-1
--- a/Makefile
+++ b/Makefile
@@ -103,11 +103,12 @@ makemigrations: ##  Run makemigrations on the docker container set MIGRATE=false
 	fi
 
 ruff: install ##  **Experimental** Run ruff linter. To fix and format files.
+	@echo "Running Mypy..."
+	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
 	@echo "Running Ruff..."
 	$(POETRY) run ruff check
 	$(POETRY) run ruff format
-	@echo "Running Mypy..."
-	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
+	
 
 
 debug: ##  Connect to the shepherd container with docker debug.
--- a/Makefile
+++ b/Makefile
@@ -104,10 +104,11 @@ makemigrations: ##  Run makemigrations on the docker container set MIGRATE=false
 
 ruff: install ##  **Experimental** Run ruff linter. To fix and format files.
 	@echo "Running Ruff..."
-	$(POETRY) run ruff check --select I --fix
+	$(POETRY) run ruff check
 	$(POETRY) run ruff format
 	@echo "Running Mypy..."
 	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
+
 
 debug: ##  Connect to the shepherd container with docker debug.
 	docker debug consvc-shepherd-app-1
--- a/ruff.toml
+++ b/ruff.toml
@@ -33,6 +33,7 @@ exclude = [
     "venv",
     "migrations",
     "manage.py",
+    "*test_*.py",
 ]
 
 [lint]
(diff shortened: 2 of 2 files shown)


kauana:
Can confirm that the only difference between running `make lint-fix` and `make ruff` is that ruff only formats lines to 120 characters which is so much better imo  🎉 . Ex:
```
-                        fill_text_field_by_id(
-                            f"id_ad_urls-{i}-domain", url_info["host"], driver
-                        )
-                        fill_text_field_by_id(
-                            f"id_ad_urls-{i}-path", path["value"], driver
-                        )
+                        fill_text_field_by_id(f"id_ad_urls-{i}-domain", url_info["host"], driver)
+                        fill_text_field_by_id(f"id_ad_urls-{i}-path", path["value"], driver)
```

nanagb:
> Can confirm that the only difference between running `make lint-fix` and `make ruff` is that ruff only formats lines to 120 characters which is so much better imo 🎉 . Ex:
> 
I like the new line formatting a lot better too

Makefile
```diff
@@ -103,8 +103,11 @@ makemigrations: ##  Run makemigrations on the docker container set MIGRATE=false
 	fi
 
 ruff: install ##  **Experimental** Run ruff linter. To fix and format files.
+	@echo "Running Ruff..."
 	$(POETRY) run ruff check --select I --fix
 	$(POETRY) run ruff format
+	@echo "Running Mypy..."
+	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
```
kauana, line 110: @nanagb I'm not sure `ruff` can do what `bandit` does for security scanning.

I used the code below as an example:
```
import json
import yaml
from yaml import CSafeLoader
from yaml import SafeLoader


def test_yaml_load():
    ystr = yaml.dump({'a': 1, 'b': 2, 'c': 3})
    y = yaml.load(ystr)
    yaml.dump(y)
    try:
        y = yaml.load(ystr, Loader=yaml.CSafeLoader)
    except AttributeError:
        # CSafeLoader only exists if you build yaml with LibYAML
        y = yaml.load(ystr, Loader=yaml.SafeLoader)


def test_json_load():
    # no issue should be found
    j = json.load("{}")

yaml.load("{}", Loader=yaml.Loader)

# no issue should be found
yaml.load("{}", SafeLoader)
yaml.load("{}", yaml.SafeLoader)
yaml.load("{}", CSafeLoader)
yaml.load("{}", yaml.CSafeLoader)
```

Bandit will report 2 medium securities issues and ruff will only fix the import lines:
```
 import json
+
 import yaml
-from yaml import CSafeLoader
-from yaml import SafeLoader
+from yaml import CSafeLoader, SafeLoader
 
 
 def test_yaml_load():
-    ystr = yaml.dump({'a': 1, 'b': 2, 'c': 3})
+    ystr = yaml.dump({"a": 1, "b": 2, "c": 3})
     y = yaml.load(ystr)
     yaml.dump(y)
     try:
@@ -19,6 +19,7 @@ def test_json_load():
     # no issue should be found
     j = json.load("{}")
 
+
 yaml.load("{}", Loader=yaml.Loader)
 
 # no issue should be found
 ```

Comment (nanagb) [reply Makefile:110]: @kauana Yes great catch! Looks like we still need bandit in addition to mypy.

Comment (nanagb) [reply Makefile]: I made some tweaks to the ruff command and it's now running the bandit tests

Comment (kauana) [reply Makefile]: Hmm I do see that `make ruff` is reporting the security issues, but it's also reporting some new things as well such as:
```

consvc_shepherd/version.py:19:35: B008 Do not perform function call `pathlib.Path.cwd` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable
   |
18 | def fetch_app_version_from_file(
19 |     app_root_path: pathlib.Path = pathlib.Path.cwd(),
   |                                   ^^^^^^^^^^^^^^^^^^ B008
20 | ) -> Version:
21 |     """Fetch the content of the version.json file, which contains the SHA-1 hash
   |
```

Maybe this is fine if we fix all the issues and start using ruff going forward though.

nanagb: Yes, these tests were not being evaluated with the way my old ruff command was formatted.

I think it makes sense to slowly transition to ruff as there may be tweaks and changes others would like to see as they test it vs the old linters. I was thinking of doing a PR with the fixes (including reformatting the line lengths) once this gets into main.

kauana (approved):

PR #266, opened 2024-09-30
chore: update docs after GCPv2 migration

ahoneiser:


ahoneiser:
let me know what you think @mashalifshin & @cyptm :slightly_smiling_face:

mashalifshin (approved):
Thanks so much for cleaning up these docs! 🙌

PR #271, opened 2024-10-03
Gracefully handle 204 responses from /v1/tiles on Preview page

cbguder:
## Problem Statement

The /v1/tiles endpoint returns a 204 response with no body when there are no tiles available. Currently when this happens, get_amp_tiles raises an exception, and we don't display any ads (even if e.g. SPOCs might be available):

Screenshot 2024-10-03 at 2 50 12 PM

## Proposed Changes

This changes get_amp_tiles to gracefully handle a 204 response from /v1/tiles and allows the preview page to display other ads that migth be available.

## Verification Steps

This can be tricky to verify, as tiles _should_ almost always available for the countries/regions currently supported by the preview page. However, at time of writing, they are not available for Canada, as seen in the above screenshot.

To verify:
1. Go to the Preview page
2. Pick a country/region where tiles are not available, and hit Preview
3. Observe that SPOCs are shown

## Visuals

Screenshot 2024-10-03 at 2 50 27 PM

### DIFF
--- a/consvc_shepherd/preview.py
+++ b/consvc_shepherd/preview.py
@@ -314,6 +314,10 @@ def get_amp_tiles(
     r = requests.get(
         f"{env.mars_url}/v1/tiles", params=params, headers=headers, timeout=30
     )
+
+    if r.status_code == 204:
+        return []
+
     return [
         Tile(
             image_url=tile["image_url"],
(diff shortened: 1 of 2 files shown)


nanagb (approved):
LGTM: I verified this was working as expected on local dev with Canada since that's currently returning a 204.

PR #274, opened 2024-10-08
GHA maintenance

ahoneiser:


mashalifshin (approved):
Thank you for doing this! I've been starting to think about building and serving the dashboard for production, this looks like a great first step. 

FYI We aren't yet linting the ad-ops-dashboard in CI/CD (gotta clean up some lint first before we can activate, but soon!), but we are testing it as part of the `make test` command.

I'm curious for my learning what prompted you to add this change now, was there an error that occurred? Or were you seeing the workflow runs get slow due to the lack of cached node dependencies?

PR #275, opened 2024-10-08
[SHEP-109] Fixed Linting Issues and Enabled Linting in Makefile

akhan-mozilla:
## References

[SHEP-123](https://mozilla-hub.atlassian.net/browse/SHEP-123)

## Problem Statement

Lint issues need to be fixed for the campaigns pages and components since the lint setup has been completed and merged.

## Proposed Changes

1. Added lint fixes for the campaigns pages and their components
2. Enabled ESLint in the `make lint` command

## Verification Steps

run `make lint` command 

[SHEP-123]: https://mozilla-hub.atlassian.net/browse/SHEP-123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ad-ops-dashboard/eslint.config.js
+++ b/ad-ops-dashboard/eslint.config.js
@@ -43,6 +43,7 @@ export default tseslint.config(
       ],
       ...react.configs.recommended.rules,
       ...react.configs["jsx-runtime"].rules,
+      "@typescript-eslint/no-misused-promises": "off",
     },
   },
 );
--- a/ad-ops-dashboard/eslint.config.js
+++ b/ad-ops-dashboard/eslint.config.js
@@ -43,7 +43,12 @@ export default tseslint.config(
       ],
       ...react.configs.recommended.rules,
       ...react.configs["jsx-runtime"].rules,
-      "@typescript-eslint/no-misused-promises": "off",
+      "@typescript-eslint/no-misused-promises": [
+        "error",
+        {
+          "checksVoidReturn": false
+        },
+      ],
     },
   },
 );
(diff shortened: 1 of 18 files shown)


ad-ops-dashboard/eslint.config.js
```diff
@@ -43,6 +43,7 @@ export default tseslint.config(
       ],
       ...react.configs.recommended.rules,
       ...react.configs["jsx-runtime"].rules,
+      "@typescript-eslint/no-misused-promises": "off",
```
kauana, line 46: Hey Ayesha, thanks for working on these changes. Could you please explain why we are turning off the `no-misused-promises` rule?

akhan-mozilla: The error regarding a `Promise-returning function being provided to an attribute where a void return was expected ` occurs in the `onSubmitHandler` function in forms. Since we have [functions](https://github.com/mozilla-services/consvc-shepherd/blob/786392abea5d73fc49026ba17183a97ade58d146/ad-ops-dashboard/src/components/Forms/SplitCampaignForm.tsx#L58) that need to return a Promise, I set `checksVoidReturn` to `false` to avoid conflicts with our logic. If I had not done this, I would have had to force the `onSubmitHandler` to return void, which is not the default behavior of our functions.

kauana: Thanks for explaining. My only concern with disabling this rule is that we might miss some helpful linting tips, as shown [here](https://typescript-eslint.io/rules/no-misused-promises/#checksvoidreturn-1). However, since changing onSubmitHandler would require a refactor, which doesn’t seem worth it, I see why that was the choice. Also, I noticed that instead of completely disabling the rule, you modified the config to specifically disable checks for promises with a void return type, so I think we’re fine.

kauana (approved):
Just ran eslint and our tests and didn't see any linting issues anymore and the tests are passing. I've also built ad-ops-dashboard and confirmed Campaign behavior (create, split, delete, update, etc) is still working. LGTM!

PR #277, opened 2024-10-10
[SHEP-70]: Add flight_name to BQ query in sync_bq_script

kauana:
## References

[SHEP-70](https://mozilla-hub.atlassian.net/browse/SHEP-70)

## Problem Statement

We received feedback from the Ad Ops folks that it would be good to see the flight name in the Delivered Flights table as well. This PR includes the flight_name to the BQ query in the sync_bq_script.

## Proposed Changes

- Update `sync_bq_data` script to also query for the flight name
- Update `test_sync_bq_data` tests to also include and test the flight name

## Verification Steps

1. Start Shepherd with `docker compose up --build`
1. Pass your gcloud credentials to the shepherd container to access the BigQuery view `moz-fx-data-shared-prod.ads.consolidated_ad_metrics_daily_pt`:
    1. Login to gcloud locally `gcloud auth application-default login`.
    2. Copy your gcloud creds into the shepherd container:
   ```bash
   docker cp ~/.config/gcloud/application_default_credentials.json consvc-shepherd-app-1:/app/application_default_credentials.json
   ```
1. Exec into the Shepherd container with `make debug`.
1. Tell gcloud to use the creds with:
    ```bash
    export GOOGLE_APPLICATION_CREDENTIALS="/app/application_default_credentials.json"
    ```
1. Run the script
    ```bash
    python manage.py sync_bq_data --project_id "moz-fx-ads-nonprod"
    ```
    If successful you should see: `BigQuery fetch completed for data <today's date>`.
1. Now that the `consvc_shepherd_deliveredflight` table was populated with flight info, check that the flights also show the flight name in the Delivered Flights UI: http://0.0.0.0:7001/admin/consvc_shepherd/deliveredflight/

[SHEP-70]: https://mozilla-hub.atlassian.net/browse/SHEP-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,6 @@ book/
 mermaid-init.js
 mermaid.min.js
 lcov.info
+
+# ADC
+application_default_credentials.json
--- a/consvc_shepherd/management/commands/sync_bq_data.py
+++ b/consvc_shepherd/management/commands/sync_bq_data.py
@@ -77,6 +77,7 @@ class BQSyncer:
                 campaign_id,
                 campaign_name,
                 flight_id,
+                flight_name,
                 provider,
                 SUM(clicks) AS clicks,
                 SUM(impressions) AS impressions
@@ -91,6 +92,7 @@ class BQSyncer:
                 campaign_id,
                 campaign_name,
                 flight_id,
+                flight_name,
                 provider
         """
 
@@ -125,6 +127,7 @@ class BQSyncer:
             campaign_id = row["campaign_id"]
             campaign_name = row["campaign_name"]
             flight_id = row["flight_id"]
+            flight_name = row["flight_name"]
             provider = row["provider"]
             clicks = row["clicks"]
             impressions = row["impressions"]
@@ -134,6 +137,7 @@ class BQSyncer:
                 campaign_id=campaign_id,
                 campaign_name=campaign_name,
                 flight_id=flight_id,
+                flight_name=flight_name,
                 provider=provider,
                 # If script runs again in the same day, just update the clicks and impressions
                 defaults={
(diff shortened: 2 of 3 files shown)


nanagb:
🎉 I got this working for today and yesterday.

But got the following error when I tried 2024-10-01
`django.db.utils.IntegrityError: null value in column "flight_name" of relation "consvc_shepherd_deliveredflight" violates not-null constraint`

This is probably because the flight name wasn't retro added to existing rows generated before `filight_name` was added to BQ?

Maybe data science can easily add the missing information or we can get it from the API. Thoughts?

nanagb:
👀 application_default_credentials.json should be added to .gitignore

kauana:
> 👀 application_default_credentials.json should be added to .gitignore

Good point! It's always better to be safe than sorry! 😅

kauana:
> 🎉 I got this working for today and yesterday.
> 
> But got the following error when I tried 2024-10-01 `django.db.utils.IntegrityError: null value in column "flight_name" of relation "consvc_shepherd_deliveredflight" violates not-null constraint`
> 
> This is probably because the flight name wasn't retro added to existing rows generated before `filight_name` was added to BQ?
> 
> Maybe data science can easily add the missing information or we can get it from the API. Thoughts?

Thank you for pointing this out. I thought this was okay because we haven't been running the sync_bq_data script yet as I was waiting for this field to be added. What do you think?

Regardless, it doesn't hurt to ask the data engineering folks if we can add the data retroactively."

nanagb:
> Thank you for pointing this out. I thought this was okay because we haven't been running the sync_bq_data script yet as I was waiting for this field to be added. What do you think?
> 
> Regardless, it doesn't hurt to ask the data engineering folks if we can add the data retroactively."

It shouldn't be an issue if we never sync in data from before the day the field was added. But I have a feeling we'll need earlier data.

Yes, let's see what data engineering says about this.

nanagb (approved):
Approving because it's working and once data eng makes the changes, past data will sync with no issue.

PR #281, opened 2024-10-11
[SHEP-92] csrf authentication error fix

akhan-mozilla:
## References

[SHEP-92](https://mozilla-hub.atlassian.net/browse/SHEP-92)

## Problem Statement

CSRF authentication needs to be added for the APIs, as Django returns CSRF authentication errors on POST, PATCH, and DELETE requests.

## Proposed Changes

Added the CSRF Origins array.
Made the authentication classes empty to resolve the CSRF authentication issue since Open IDC authentication is applied.

## Verification Steps

You can test the fix by removing the empty authentication class to check csrf errors. To find out how Open IDC authentication conflicts with Django authentication, we can remove the OpenIDC authentication and use the default Django authentication classes. This should not return any CSRF-related errors.

[SHEP-92]: https://mozilla-hub.atlassian.net/browse/SHEP-92?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/consvc_shepherd/settings.py
+++ b/consvc_shepherd/settings.py
@@ -261,7 +261,7 @@ markus.configure(backends=_MARKUS_BACKENDS)
 
 CORS_ALLOWED_ORIGINS = ["http://0.0.0.0:5173", "http://localhost:5173"]
 
-CSRF_TRUSTED_ORIGINS = ["http://localhost:5173"]
+CSRF_TRUSTED_ORIGINS = ["http://0.0.0.0:5173", "http://localhost:5173"]
 
 REST_FRAMEWORK: Dict[str, Any] = {
     "DEFAULT_AUTHENTICATION_CLASSES": [],
--- a/consvc_shepherd/settings.py
+++ b/consvc_shepherd/settings.py
@@ -2,7 +2,7 @@
 
 import sys
 from pathlib import Path
-from typing import Any
+from typing import Any, Dict
 
 import environ
 import markus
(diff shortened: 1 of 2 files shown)


akhan-mozilla:
We are using OpenIDC for API authentication, which alters the default Django authentication classes and leads to CSRF errors due to the unhandled sessionAuthentication class. I made the default Django authentication empty to avoid CSRF errors related to the sessionAuthentication class since we are not using the session authentication. I added the CSRF Trusted Origins class to allow the addition of the React app URL as a trusted origin.

consvc_shepherd/settings.py
```diff
@@ -259,6 +259,10 @@
...
-CORS_ALLOWED_ORIGINS = [
-    "http://0.0.0.0:5173",
-]
+CORS_ALLOWED_ORIGINS = ["http://0.0.0.0:5173", "http://localhost:5173"]
+
+CSRF_TRUSTED_ORIGINS = ["http://localhost:5173"]
+
+REST_FRAMEWORK: Dict[str, Any] = {
+    "DEFAULT_AUTHENTICATION_CLASSES": [],
+}
```
kauana, line 268: Thanks for the PR! I can confirm I'm no longer seeing those 403 errors I've encountered before. 

These changes are good for local development. I'm wondering if we have a story in the backlog to update these permissions when we release the app to stage/prod?

consvc_shepherd/settings.py
```diff
@@ -259,6 +259,10 @@
     )
 markus.configure(backends=_MARKUS_BACKENDS)
 
-CORS_ALLOWED_ORIGINS = [
-    "http://0.0.0.0:5173",
-]
+CORS_ALLOWED_ORIGINS = ["http://0.0.0.0:5173", "http://localhost:5173"]
+
+CSRF_TRUSTED_ORIGINS = ["http://localhost:5173"]
```
kauana, line 264: Should we also add `http://0.0.0.0:5173` to the list of CSRF trusted origins?

consvc_shepherd/settings.py
akhan-mozilla: Sure, I've created a [ticket](https://mozilla-hub.atlassian.net/browse/SHEP-128) to update these permissions when we release the app to staging/prod

consvc_shepherd/settings.py
akhan-mozilla: Good catch! I've just added the `http://0.0.0.0:5173 ` to CSRF_TRUSTED_ORIGINS list."

kauana (approved):
LGTM 🚢

PR #282, opened 2024-10-14
[SHEP-125] - make campaign flight names nullable

kauana:
## References

[SHEP-125](https://mozilla-hub.atlassian.net/browse/SHEP-125)

## Problem Statement

We’ve learned from Curtis that there might be some special scenarios where the flight_name will be missing in some flights. Let’s change our db to allow flight names to be null as long as the flight_id is not null so we can actually aggregate click and impression data.

Let’s also do the same for campaign id just in case they might be missing in some flights, making sure that campaign_id can not be null.

## Proposed Changes

- Allow `flight_name` and `campaign_name` to be null
- Bug fix: Remove `flight_name` and `campaign_name` from DB lookup condition
- Add unit tests focused on flight name and campaign name being null and running the script more than once for the same day
- Enhance previous tests

```
This is a fix for a bug we were having when running the script again in the
same day. The problem was that when we ran it twice in the same day we would
get the error: `CommandError: duplicate key value violates unique constraint
"unique_delivered_flight"`.

This issue is happening because campaign_name and flight_name are being
included in the lookup condition. As a result, update_or_create is searching
for an exact match on campaign name, flight name, in addition to our unique
  constraints submission date, campaign id, flight id, and provider. When it
  doesn't find a matching record (such as when those non-nullable fields aren't
  an exact match), it attempts to create a new entry, but there's already a
  record with the same unique constraints combination, leading to the duplicate
  key error.

By moving the campaign name and flight name fields into defaults, they are only
updated or set and not used in the lookup.
```

## Verification Steps

- Run shepherd and exec with `make debug`
Run the script with `python manage.py sync_bq_data --project_id "moz-fx-ads-nonprod" --date 2024-10-01` and ensure that it succeeds.
    Note: before this PR, we would get error:
    ```
    CommandError: An error occurred: null value in column "flight_name" of relation "consvc_shepherd_deliveredflight" violates not-null constraint
    DETAIL:  Failing row contains (2306, 2024-10-01, <campaign id>, <campaign name>, <flight_id>, null, kevel, 0, 2).
    
    In this case, the flight_name is null.
    ````

- Run the script again for the same date and check that it succeeds.
    Note: before this PR, running the script again failed with the error: 
    ```
    CommandError: duplicate key value violates unique constraint "unique_delivered_flight"
    DETAIL:  Key (submission_date, campaign_id, flight_id, provider)=(2024-10-10, <campaign_id>, <flight_id>, kevel) already exists.
    ```

[SHEP-125]: https://mozilla-hub.atlassian.net/browse/SHEP-125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nanagb (approved):
New code working. No more errors when the flight_name is null. Also confirmed that it updates data on exiting rows if the script is run again with the same date range. 
Thanks for all the extra unit tests.

PR #284, opened 2024-10-14
[SHEP-17] - Add a default project-id to BQ sync script

kauana:
## References

[SHEP-17](https://mozilla-hub.atlassian.net/browse/SHEP-17)

## Problem Statement

[Andre suggested](https://github.com/mozilla-it/webservices-infra/pull/2900#discussion_r1797262801) it would be easier to configure the infra side if the script handled the project-id logic (instead of configuring the k8s job to use that info).

## Proposed Changes
- Use  a default project id `moz-fx-ads-prod` to query BQ
- Add tests for the default project id

## Verification Steps

- Run shepherd and exec with `make debug`
- Run the script without providing a project ID with  `python manage.py sync_bq_data` and check that the script runs successfully.
- Try passing a different project id with ``python manage.py sync_bq_data --project_id "moz-fx-ads-nonprod"` and check that the script runs successfully
- Run the tests

[SHEP-17]: https://mozilla-hub.atlassian.net/browse/SHEP-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/consvc_shepherd/management/commands/sync_bq_data.py
+++ b/consvc_shepherd/management/commands/sync_bq_data.py
@@ -28,7 +28,7 @@ class Command(BaseCommand):
             "--project_id",
             default=DEFAULT_PROJECT_ID,
             type=str,
-            help="The GCP project ID that will interact with BQ. By default, it will use \"moz-fx-ads-prod\"",
+            help='The GCP project ID that will interact with BQ. By default, it will use "moz-fx-ads-prod"',
         )
         parser.add_argument(
             "--date",
--- a/consvc_shepherd/management/commands/sync_bq_data.py
+++ b/consvc_shepherd/management/commands/sync_bq_data.py
@@ -14,6 +14,7 @@ from consvc_shepherd.models import BQSyncStatus, DeliveredFlight
 
 SYNC_STATUS_SUCCESS = "success"
 SYNC_STATUS_FAILURE = "failure"
+DEFAULT_PROJECT_ID = "moz-fx-ads-prod"
 
 
 class Command(BaseCommand):
@@ -25,8 +26,9 @@ class Command(BaseCommand):
         """Register expected command line arguments"""
         parser.add_argument(
             "--project_id",
+            default=DEFAULT_PROJECT_ID,
             type=str,
-            help="The GCP project ID that will interact with BQ, e.g. moz-fx-ads-nonprod.",
+            help="The GCP project ID that will interact with BQ. By default, it will use \"moz-fx-ads-prod\"",
         )
         parser.add_argument(
             "--date",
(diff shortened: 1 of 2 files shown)


mashalifshin (approved):
The change to have a default project id works, and the old behavior to pass it works as well!

Would you mind updating the PR title / commit message with something like "Add a default project-id to BQ sync script"?

I'm wondering about the choice to keep it as a command line flag, instead of passing it as an environment variable. I dipped into the discussion on the infra PR, and I don't fully understand it, but it seemed like André was recommending environment variable, so I was curious about that. 

(I know on the Boostr sync script side, we started out using command line flags, but later learned environment variables were often the better option for the cron infra setup, so we've got a mix on that script)

I'll Approve in the meantime so you can keep moving quickly on the cron job setup, but just wanted to _flag_ (har har har) this question

kauana:
> The change to have a default project id works, and the old behavior to pass it works as well!
> 
> Would you mind updating the PR title / commit message with something like "Add a default project-id to BQ sync script"?
> 
> I'm wondering about the choice to keep it as a command line flag, instead of passing it as an environment variable. I dipped into the discussion on the infra PR, and I don't fully understand it, but it seemed like André was recommending environment variable, so I was curious about that.
> 
> (I know on the Boostr sync script side, we started out using command line flags, but later learned environment variables were often the better option for the cron infra setup, so we've got a mix on that script)
> 
> I'll Approve in the meantime so you can keep moving quickly on the cron job setup, but just wanted to _flag_ (har har har) this question

Good question. I chose to set it as a default parameter instead of an environment variable because in my head the project id is something that we won't be changing much and its not something secret such as the boostr email/password. Also, I didn't want to have to set a new environment variable in `.env` for that.

I'm still allowing a custom project-id in case we would like to run the script locally like we do the boostr sync script, from time to time so in that occasion we could pass the ads nonprod project id instead.

Having said that, I'm not opposed to using an environment variable at all. If the boost sync script will be relying more on env vars, I'm happy to make this change so we can have more consistent across our scripts.

mashalifshin:
> Good question. I chose to set it as a default parameter instead of an environment variable because in my head the project id is something that we won't be changing much and its not something secret such as the boostr email/password. Also, I didn't want to have to set a new environment variable in `.env` for that.
> 
> I'm still allowing a custom project-id in case we would like to run the script locally like we do the boostr sync script, from time to time, so in that occasion we could pass the ads nonprod project id instead.
> 
> Having said that, I'm not opposed to using an environment variable at all. If the boost sync script will be relying more on env vars, I'm happy to make this change so we can have more consistent across our scripts.

The boostr sync script also still uses some parameters, so I think from the perspective of consistency it'd be a fine choice to leave as-is. And your reasoning re. it not being a secret makes sense in that it doesn't _need_ to be an environment variable necessarily. I don't feel like I fully understand what are the infra implications to this choice, so if you want to be super diligent perhaps you can ask André on the other PR, if it makes a substantial difference from the infra perspective whether we use the command line arg or the environment variable for this.

Totally okay with merging this as-is and iterating on it in subsequent PRs if the infra considerations make it worthwhile

PR #285, opened 2024-10-14
[SHEP-72] Added flights table to save flights for campaigns

akhan-mozilla:
## References

[SHEP-72](https://mozilla-hub.atlassian.net/browse/SHEP-72)

## Problem Statement

Campaigns can have multiple associated flights. To accommodate this, we need to create a separate table for flights.

## Proposed Changes

1. Added a new table for flights. 
2. Updated the existing campaign functionality to utilize the newly created flights table for updating or creating flights associated with campaigns.

## Verification Steps

Create or update campaigns by adding their flight IDs and verify that these are saved in the Flights table for accurate campaigns.

## Visuals

Screenshot 2024-10-14 at 2 18 36 PM

[SHEP-72]: https://mozilla-hub.atlassian.net/browse/SHEP-72?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


consvc_shepherd/api/serializers.py
```diff
@@ -21,13 +21,20 @@ class CampaignSerializer(serializers.ModelSerializer):
...
     )
+    kevel_flight_id = serializers.IntegerField(write_only=True, allow_null=True)
 
     class Meta:
         """Metadata to specify the way Campaign is serialized"""
 
         model = Campaign
         fields = "__all__"
 
+    # To do: This code will be updated once we have the UI to assign multiple flights to a campaign.
```
mashalifshin, line 32: Thanks for the note, that makes sense. You don't want to change the API yet and break the React side. Feel free to make an additional ticket for this unless it's already captured.

consvc_shepherd/migrations/0023_remove_campaign_kevel_flight_id_flight.py
```diff
@@ -0,0 +1,46 @@
...
+                (
+                    "id",
+                    models.BigAutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("flight_id", models.IntegerField(blank=True, null=True)),
```
mashalifshin, line 30: Two suggestions here
- Could we please call this `kevel_flight_id` as before? We want to be really explicit that this id must be from kevel.
- Let's add a `name` column on the flights table, I think it could get really hard to manage otherwise for the case of multiple flights (in [Laura's diagram](https://docs.google.com/presentation/d/16eiVkDSR9tILspLQMMV6ngWiMzUKBlxwlaV0m5r1_4A/edit#slide=id.g30296749286_0_0) they have names).

And one more idea, out of scope for this PR, but we could make an enhancement ticket to also populate a name like "March Tiles" into that field, like we see on the diagram, since we can derive that from the campaign's dates and the product

consvc_shepherd/models.py
```diff
@@ -460,6 +457,13 @@ def net_ecpm(self):
             return round(net_epcm_value, 2)
         return None
 
+    # To do: This code will be updated once we have the UI to assign multiple flights to a campaign.
+    @property
+    def kevel_flight_id(self):
```
mashalifshin, line 462: Just making sure I'm understanding correctly, but in that next iteration, this will become the property `kevel_flight_ids` and return an array of integers, right?

consvc_shepherd/models.py
```diff
@@ -442,7 +440,6 @@ class Campaign(models.Model):
 
     ad_ops_person: CharField = models.CharField(null=True, blank=True)
     notes: CharField = models.CharField(null=True, blank=True)
-    kevel_flight_id: IntegerField = models.IntegerField(null=True, blank=True)
```
mashalifshin, line 445: When I look at the [Campaigns list view on your branch](http://localhost:7001/admin/consvc_shepherd/campaign/), I'm still seeing `kevel_flight_id`. Could we adjust that to show `kevel_flight_ids` and do a comma separated list of all the ids?

mashalifshin: Also it would be great to have the [Campaigns edit view](http://localhost:7001/admin/consvc_shepherd/campaign/2/change/) list the flights and have them be changeable there as well. iirc it's just a bit more config on the admin.

(There is an example of this when viewing the `BoostrDeal`, we also list the children in that view, Django calls it an "inline")

consvc_shepherd/api/serializers.py
```diff
@@ -54,25 +61,59 @@ def validate(self, data):
...
-            except Campaign.DoesNotExist:
-                continue
+        self._update_existing_campaigns(campaign_fields_data)
+
+        if kevel_flight_id is not None:
+            Flight.objects.create(campaign=campaign, flight_id=kevel_flight_id)
 
         return campaign
 
+    def update(self, instance, validated_data):
```
mashalifshin, line 75: Please lmk what I might be missing here, but I'm not sure why we're handling each field of the `validated_data` separately. Would it be possible to simplify this code with the use of `Campaign.update_or_create`? You can pass whole objects in to create the associations you need 
(like what we do in boostr sync script [here](https://github.com/mozilla-services/consvc-shepherd/blob/main/consvc_shepherd/management/commands/sync_boostr_data.py#L323)?

mashalifshin (commented):
Hi Ayesha, thanks for taking on the schema refactor! 💪 🧠 ❤️ 

Just a couple schema tweaks requested here, and some additional (hopefully small) updates to the admin UI

consvc_shepherd/models.py
akhan-mozilla: Yes, that's right.

mashalifshin: Gotcha, makes sense, thanks for confirming!

consvc_shepherd/migrations/0023_remove_campaign_kevel_flight_id_flight.py
mashalifshin: @akhan-mozilla apologies for sending you down the wrong path, but I actually think you should disregard adding `name` here. Just realized we are actually getting `flight_name` from BQ and it will be in the `DeliveredFlight` table, so I think we should be grabbing it from there instead of having a field for it here. Lmk if that sounds good, and thanks for bearing with me

Screenshot for reference

Screenshot 2024-09-30 at 3 54 38 PM

consvc_shepherd/api/serializers.py
akhan-mozilla: Yes, I have the ticket [SHEP-126](https://mozilla-hub.atlassian.net/browse/SHEP-126) to update the UI to add multiple flights for campaigns, and I'll make these changes as part of that.

consvc_shepherd/migrations/0023_remove_campaign_kevel_flight_id_flight.py
akhan-mozilla: Yes, it sounds good to use the `flight_name` from `DeliveredFlights`.

consvc_shepherd/models.py
akhan-mozilla: I did not make changes to the Admin UI for the Campaigns list and edit pages because updating the `kevel_flight_id` property to return an array and saving multiple flights for campaigns would disrupt the current behavior of the Campaigns API and break the frontend.
I will address these points as part of the multiple flights [tickets](https://mozilla-hub.atlassian.net/browse/SHEP-126), ensuring that multiple flights are correctly assigned and displayed in both the Django admin and React UI. I've also noted these points in [SHEP-126 ](https://mozilla-hub.atlassian.net/browse/SHEP-126)to track them.

consvc_shepherd/api/serializers.py
akhan-mozilla: The `update` method I added here is the default serializer method, which is automatically called with the PATCH API. However, the `update_or_create` method is specific to Django ORM queries and cannot be used in this context. I updated the default Django serializer’s `update` method because the campaign data is not saved in a single table, and to ensure that individual flights for campaigns are saved correctly in the Flight table.

consvc_shepherd/models.py
mashalifshin: Oh right, thank you so much for explaining that, totally makes sense. Appreciate you tracking these in the ticket and will look forward to the subsequent PRs 😁

consvc_shepherd/api/serializers.py
mashalifshin: Ahh thanks for the background on needing to create our own version of the `update` method instead of using the default for the serializer. Continuing to learn my way around the standard Django stuff here, so I appreciate the explanation!

I think you miiiiiight still be able to use a single `Campaign.objects.update_or_create` call instead of the separate calls to update the Flight on line 86 and to update the Campaign on line 107, saving a trip to the DB for the separate calls, and another trip to the DB to check if the Campaign exists. (I can tell those ORM methods are available because you are using one of them on that same line 86, for the`Flight` model). But that's totally not a merge blocker, just an optimization thought.

mashalifshin (approved):
Ayesha I took it for a spin and everything looks great! Campaigns form (create and update cases) is still working like a charm with this refactor. 

I think you might be able to optimize a bit more in that `update` serializer function, but definitely not required for merge.

akhan-mozilla:
> Ayesha I took it for a spin and everything looks great! Campaigns form (create and update cases) is still working like a charm with this refactor.
> 
> I think you might be able to optimize a bit more in that `update` serializer function, but definitely not required for merge.

Thanks for pointing this out. I'll do these optimizations for `update_or_create` as part of my next PR.

consvc_shepherd/api/serializers.py
akhan-mozilla: Thanks for pointing this out. I'll do these optimizations for update_or_create as part of my next PR.

mashalifshin: Ayesha thanks for always being so willing to refactor! I slept on it and I think you're right we do need separate calls for the Flight and the Campaign, but also still think you might be able to use update_or_create for both and save the call to check if the Campaign exists ... anyway, I look forward to what you come up with next! <3

PR #287, opened 2024-10-16
[SHEP-36]: Adding make command to show human readable timestamps in local docker logs

luc-lisi:
## References

[SHEP-36](https://mozilla-hub.atlassian.net/browse/SHEP-36)

## Problem Statement

The original ask was to modify the timestamp in the log outputs to display in a more human readable format to allow for easier local debugging as by default, docker logs in the command line do not show human readable timestamps. (They do in Docker Desktop however, but this is not as convenient.)

## Proposed Changes

The [JSON formatting library](https://python-dockerflow.readthedocs.io/en/main/api/logging.html) we are using for logging conforms to the [MozLog schema](https://wiki.mozilla.org/Firefox/Services/Logging). As a result, we cannot (and probably shouldn't) modify the timestamp field to be in a different format.

However, to achieve the same goal of making local dev logs more easy to parse, we can make some changes to how we run our docker containers and add a new make command `dev-ts` that will display timestamps next to all docker logs. This however does come with a downside that we have to run `docker-compose up` in detached mode, meaning the usual `ctrl-c` motion will simply close out of the log follower, not kill the containers. 

In addition, for scripts run through manage.py from inside the docker container we've included the option for setting a local env variable which will append human readable datetime headers in front of each log. To do this the following changes were made:

 1. Added a new (optional) env variable APP_ENV. This may be useful in the future for other things as well and we may want to require it down the line, but for now it can be omitted without issue.
 2. Added a datetime header in front of all logs emitted when running scripts via manage.py. These headers will only be included if APP_ENV == "dev" and previous behavior (with no datetime headers) is expected with any other value for APP_ENV or if it is undefined.

Finally, we also updated the docs to add info about the schema of our Json logs.

## Verification Steps
1.  **Verify console docker timestamps** 
- [ ] run `make dev-ts`
- [ ] Verify human-readable timestamps are visible next to logs.

2. **Verify default behavior in scripts is no datetime headers on logs**
- [ ] verify no `.env` variable exists named `APP_ENV`
- [ ] run `python manage.py sync_boostr_data https://app.boostr.com/api/` inside the docker container
- [ ] verify logs look as they did before this PR (no datetime headers)

3. **Verify new behavior in scripts is _with_ datetime headers on logs**
- [ ] add a new `.env` variable with `APP_ENV=dev`
- [ ] run `python manage.py sync_boostr_data https://app.boostr.com/api/` inside the docker container
- [ ] verify logs now have datetime headers in a newline above each log.

## Visuals
New behavior with `make dev-ts` command.
Screenshot 2024-10-15 at 5 55 05 PM

New behavior in scripts with `APP_ENV=dev`
Screenshot 2024-10-16 at 1 10 37 PM

[SHEP-36]: https://mozilla-hub.atlassian.net/browse/SHEP-36?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kauana:
Hey @luc-lisi,

Thank you for adding the human-readable timestamps to our Docker logs! I really like this a lot 🥳.

However, I believe the deliverable for SHEP-36 is to actually format the timestamps in our `sync_boostr_data` script. We have some documentation [here](https://github.com/mozilla-services/consvc-shepherd/blob/main/docs/operations/syncBoostrData.md) on how to run the script. Once you run it, you'll see logs similar to the image shared in the Jira ticket, where we have Unix epoch timestamps that look like `"Timestamp": 1729094620695415296`. These are the timestamps we want to convert to human-readable format.

That said, I do like your change, and we should definitely keep it as well.

luc-lisi:
Hey @kauana! 

Understood! Yeah it might be good to get some more context on this ticket then. What is the purpose for changing the timestamp from UTC nano to a human-readable format? Speaking to @mashalifshin it sounded like maybe this was a convince thing for local dev, but if a service downstream is requiring this change than we can definitely add a human readable section to make it work.

For context, currently it looks like our logging is build on top of dockerflow's `JsonLogFormatter` library which strictly adheres to the [MozLog JSON schema](https://wiki.mozilla.org/Firefox/Services/Logging). In these docs its seems pretty explicit that the `Timestamp` field should always be UTC Nanoseconds but does say "If human readable time is absolutely necessary, add an additional "Time" field (RFC3339)."

So in order to do this I guess we have a few options that I see but open to other suggestions!
1. Move away from the dockerflow JsonLogFormatter we've been using and use our own custom JSON schema (not sure if this will break down stream log ingestion?)
2. Add a new Time field with human readable time (this won't be the first field however in the log and will also not be a top level prop. e.g. `{ "Timestamp": 1519361465000000000 .... "Fields":{..."Time": YYYY-MM-DDTHH:MM:SS.sss+ZZ:ZZ}}`)

Let me know your thoughts!

luc-lisi:
Actually let me double check something real quick. Maybe I'm misunderstanding this ticket.

kauana:
> Hey @kauana!
> 
> Understood! Yeah it might be good to get some more context on this ticket then. What is the purpose for changing the timestamp from UTC nano to a human-readable format? Speaking to @mashalifshin it sounded like maybe this was a convince thing for local dev, but if a service downstream is requiring this change than we can definitely add a human readable section to make it work.
> 
> For context, currently it looks like our logging is build on top of dockerflow's `JsonLogFormatter` library which strictly adheres to the [MozLog JSON schema](https://wiki.mozilla.org/Firefox/Services/Logging). In these docs its seems pretty explicit that the `Timestamp` field should always be UTC Nanoseconds but does say "If human readable time is absolutely necessary, add an additional "Time" field (RFC3339)."
> 
> So in order to do this I guess we have a few options that I see but open to other suggestions!
> 
>     1. Move away from the dockerflow JsonLogFormatter we've been using and use our own custom JSON schema (not sure if this will break down stream log ingestion?)
> 
>     2. Add a new Time field with human readable time (this won't be the first field however in the log and will also not be a top level prop. e.g. `{ "Timestamp": 1519361465000000000 .... "Fields":{..."Time": YYYY-MM-DDTHH:MM:SS.sss+ZZ:ZZ}}`)
> 
> 
> Let me know your thoughts!

Ahhhh, I just chatted with Masha, and it turns out this is indeed a Mozilla standard! So yeah, I don't think we should change our script's logging format, especially since we run the script in production and services may be expecting the current format.

Since we also run it locally, one idea that comes to mind is editing the script to check for an ENVIRONMENT variable that can be set to either "prod" or "dev." If it's set to "dev," we could use human-readable timestamps, and if it's set to "prod," we stick to the standard logs. I’m curious to hear what you all think and whether you’d find that helpful. Personally, I like having the timestamps in a more readable format because it helps me track how long the script takes to run or understand the Boostr API timeouts (I swear the timeouts are > 46 seconds 😂). But since I don't run the script all the time, it’s not a huge deal for me.

mashalifshin:
Great discussions here! @kauana thanks for raising this in the review, and thanks to @luc-lisi for being willing to look more closely here. It's always fun to see how deep the rabbit hole can go on what looks like a "simple quick win" 🐇 🕳️ 👋 

I'm liking the idea of using human readable timestamps depending on what environment we're running in! I think we already have something we can hook into for that, sorry I didn't think of this right away! 

Take a look at [`settings.py` line `179`](https://github.com/mozilla-services/consvc-shepherd/blob/main/consvc_shepherd/settings.py#L179) which configures the logging, looks like we're already using a different formatter in dev...

mashalifshin:
Oh and I should mention the boostr sync script uses the same logger that's configured there in `settings.py`, so what you set up here will work throughout. And agreed that I mostly need this for when I'm running the sync script and dealing with boostr API rate limits issues

luc-lisi:
Hey folks. Great discussion here and thank you all so much for helping point me in the right direction especially while I'm just getting my bearings! I just synced with @kauana and definitely was misunderstanding the ask here.

I've gone ahead and done two things:
1. Added a new (optional) env variable `APP_ENV`. This may be useful in the future for other things as well and we may want to require it down the line, but for now it can be omitted without issue.
2. Added a datetime header in front of all logs emitted when running scripts via manage.py. These headers will _only_ be included if `APP_ENV == "dev"` and previous behavior (with no datetime headers) is expected with any other value for APP_ENV or if it is undefined.
3. Updated the docs to add info about the schema of our Json logs.

This will should allow all current, and future, scripts run through manage.py to include datetime headers on logs when running in dev, but shouldn't affect any production runs of these scrips.

Let me know if this sounds like an alright solution @mashalifshin @cyptm! Also extra thank you to @kauana for talking this through on a call!

luc-lisi:
> Take a look at [`settings.py` line `179`](https://github.com/mozilla-services/consvc-shepherd/blob/main/consvc_shepherd/settings.py#L179) which configures the logging, looks like we're already using a different formatter in dev...

@mashalifshin Ah yeah this is originally where I found that we were using `dockerflow.logging.JsonLogFormatter`. I'm not sure I'm 100% following though on the different formatter for dev. I had thought this file were global settings for the django app? But I'm new to django so might be misunderstanding here!

luc-lisi:
> I thought it would be helpful to have the human readable timestamps in all environments including production to help in debugging when things go wrong. Any reservations with keeping the original format so that we don't break things and adding the human readable format? I don't have a strong opinion and can go either way

Hi @cyptm , thank you for these comments! Yeah so on most log-viewing UIs (GCP/AWS/Datadog/Etc.) will automatically show, or have an option, to show human readable timestamps for each log, so in order not to duplicate these stamps, we were going to keep these to dev only. That said, if there is any concern about how our log dashboards are showing that data that could definitely be included! I'm not super familiar with all the tools we use to view prod logs.

> Can we take this opportunity to add intro.md file to docs/SUMMARY.md

Added!

> I'd vote for using this as the main command dev so that we don't have both dev and dev-ts This is to keep things simple and and not have too many similar commands in the Makefile

Done! The only caveat now is that we will be running containers in detached mode, so `ctrl-c` will no longer close the container, just kill the log follower. But if that's alright, then we can keep this change!

Makefile
```diff
@@ -100,6 +100,9 @@ doc-preview: doc  ##  Preview Merino docs via the default browser
 dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
 	docker compose up
 
+dev-ts: $(INSTALL_STAMP)  ##  Run shepherd locally and show human readable timestamps.
```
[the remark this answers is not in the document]
luc-lisi: Fixed!

docs/intro.md
```diff
@@ -38,6 +38,9 @@ All snapshots are uploaded as a JSON file to a Google Cloud bucket.
 A pre-defined JSON schema validates the snapshot output prior to sending the snapshot to Google Cloud.
 From there, Contile accesses the settings from the uploaded JSON file to serve Firefox users the correct advertising tiles in each new tab.
 
+## Logging
```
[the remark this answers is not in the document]
luc-lisi: Done!

Makefile
```diff
@@ -97,8 +97,8 @@ doc: ##  Generate docs via mdBook
 doc-preview: doc  ##  Preview Merino docs via the default browser
 	mdbook serve --open
 
-dev: $(INSTALL_STAMP)  ##  Run shepherd locally and reload automatically
-	docker compose up
+dev: $(INSTALL_STAMP)  ## Run shepherd locally and show human readable timestamps.  
+	docker-compose up -d && docker-compose logs -f -t
```
mashalifshin, line 101: This should be `docker compose`, they updated the command so `docker-compose` doesn't work anymore for the newer versions

manage.py
```diff
@@ -1,12 +1,25 @@
...
+import logging
+import environ
+
+from pathlib import Path
 
 
 def main():
     """Run administrative tasks."""
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "consvc_shepherd.settings")
+    env = environ.Env(APP_ENV=(str, OPTIONAL))
```
mashalifshin, line 15: I'm not sure what concrete suggestion to make here since I'm still new to python/django and its envs, so definitely feel free to educate me here, but this doesn't feel great to me because

1. We already have a `SHEPHERD_ENV` environment variable, so adding an `APP_ENV` feels a little odd here because we only consume that value for deciding whether to show logs in a particular way. I wonder if we want to make it a much more specific and declarative environment variable, something like `LOCAL_DEVELOPMENT_LOGGING` getting set to true or false.

3. We're already loading the environment in the line above. Having a one-off special case in our `main` function to read the `.env` file again, just for this dev logging, doesn't feel right. I wonder if we might learn more about the `os.environ.setdefault` and if there is a more standard way to override those defaults? And more generally am wondering about a more standard way to achieve what we want here ...

My comment from before was getting at the idea that `settings.py` defines formatters and handlers for logging, so maybe we could define a different formatter (and handler?) in there, and use it just for local development? 

Or perhaps the right thing to do is to use a different `DJANGO_SETTINGS_MODULE` here, something like `consvc_shepherd_local_development.settings`...


So I know I don't have a concrete suggestion for you, but the gist of this comment is that I'd love for this to not be such a one-off, and for it to hook into the standard logging setup and the standard environment variable loading setup as much as possible ...

mashalifshin (commented):
Apologies for the late review! Hoping these comments are helpful for a potential future PR

Makefile
luc-lisi: Oh great catch! I wasn't aware of that!

PR #291, opened 2024-10-17
[SHEP-17] Update Project ID to be an environment variable instead of a flag

kauana:
## References

[SHEP-17](https://mozilla-hub.atlassian.net/browse/SHEP-17)

## Problem Statement

I learned that we run the `sync_bq_data` script on both the k8s prod and k8s stage environments, so the flag strategy with a default value doesn't work in this context because:

1. We had some issues configuring our Helm charts to set the flags.
2. The script runs in both k8s prod and k8s stage environments, so if you set a default project ID to prod, it will use the prod project in the stage environment as well, which is not ideal.

I made the flag change in an earlier PR, not realizing we ran two k8s environments (silly now that I think of it, but oh well 🤷‍♀️).

Andre recommended updating the script to read the project ID from an environment variable and configured the cronjob to use that variable [here](https://github.com/mozilla-it/webservices-infra/pull/3096), so this PR is now in sync with his changes on the infra side.

## Proposed Changes

- Update the `sync_bq_data` script to use an environment variable `PROJECT_ID`. This variable should be set in our `.env` file so Django will load the `PROJECT_ID` as an environment variable for us (defaults to `moz-fx-ads-nonprod` since we are doing local development).
- I've updated the script docs to mention that we need PROJECT_ID set before running `sync_bq_data`
- Tests are updated to use a mock PROJECT_ID
- Note that I'm still keeping `date` as a flag because I've had to run queries for different dates multiple times so it's just easier changing the command to use a different date instead of changing an env var. `date` is still an optional flag that will just default to today's date if left unset.

## Verification Steps
- Follow the steps in the [updated doc](https://github.com/mozilla-services/consvc-shepherd/blob/shep-17-update-project-id-usage/docs/operations/syncBigQueryData.md) to use a PROJECT_ID from `.env` file. The script should run successfully and you should see new data show up in DeliveredFlights table.

[SHEP-17]: https://mozilla-hub.atlassian.net/browse/SHEP-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nanagb (approved):
Tested and working on local.

PR #295, opened 2024-10-21
[SHEP-143] Unit test for Unmanaged Views

nanagb:
## References

[SHEP-143](https://mozilla-hub.atlassian.net/browse/SHEP-143)

## Problem Statement

Our current views are manually generated and don't have a method to enforce alignment with the models that back the views. This leads to errors when the view's SQL is updated with different fields that what was defined in the model.

## Proposed Changes

Add a new unit test that scans the database for all unmanaged models and ensures that the views that we manually create match up to the model definitions.

## Verification Steps

You can test this out in 2 ways. 
1. Edit the the CampaignSummary model without editing the view. This should fail the test
2. Edit the campaign_summary_view in the migration 0028 without updating the model. This will also fail.

Note: To test this quickly without errors add `DROP VIEW IF EXISTS ... ` to the migration file, then edit the view SQL. I've included a simple change as an example
```
    migrations.RunSQL(
            """
            DROP VIEW IF EXISTS campaign_summary_view;
            CREATE OR REPLACE VIEW campaign_summary_view AS
            SELECT
                d.id AS deal_id,
                d.id AS deal_id2,
                ...
```
               

## Visuals
Example error when the view is missing a column defined in the model
image

Example error when the view doesn't match the model because it has extra columns that don't exist in the model
image

[SHEP-143]: https://mozilla-hub.atlassian.net/browse/SHEP-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kauana:
@nanagb the first verification step says:
> Edit the the CampaignSummary model without editing the view. This should fail the test

I've modified CampaignSummary to have unique values with `impressions_sold: FloatField = models.FloatField(unique=True)` but I run the tests with `make local-test-django` and they pass. Am I missing something?

I've checked and the `test_database_views` does run:
```
app-1  | consvc_shepherd/tests/test_database_views.py                                                            23      2    91%   27-28
```

kauana:
Also, could you please update the PR title to include brackets around the ticket ID for consistency: `[SHEP-143] Unit test for unmanaged views`?

consvc_shepherd/tests/test_database_views.py
```diff
@@ -0,0 +1,52 @@
...
+        with connection.cursor() as cursor:
+            try:
+                actual_columns = {
+                    col.name
+                    for col in connection.introspection.get_table_description(
+                        cursor, view_name
+                    )
+                }
+            except Exception as e:
+                self.fail(f"Failed to get view columns for '{view_name}': {e}")
```
kauana, line 28: I see. Making fields unique or forcing them to be not null (i.e. removing `null=True, blank=True`) doesn't make the test fail because we aren't changing the columns names and instead we are changing the column constraints.

consvc_shepherd/tests/test_database_views.py
```diff
@@ -0,0 +1,52 @@
+"""Unit tests for all the database views"""
+
+from django.apps import apps
+from django.db import connection
+from django.test import TestCase
+
+
+class ViewModelAlignmentTest(TestCase):
+    """Unit tests for each view. The views are automatically detected using apps.get_models()
+    and fitering for models that are not managed.
```
kauana, line 10: typo: `fitering`

nanagb: ty fixed

consvc_shepherd/tests/test_database_views.py
nanagb: Yes, this only focuses on ensuring there's a match with the columns created by the view and the columns defined in the model. Unless, I'm overlooking something, changes in the model constraints shouldn't affect the view.

kauana: I think we should have some sort of test for constraints changes, if possible, because they can affect the views. In fact, when we had that prod issue it happened because I accidentally made `kevel_flight_id` unique 🤦‍♀️  and the database wasn't expecting it (see [this issue](https://mozilla-hub.atlassian.net/browse/SHEP-96) for more context).

kauana: I tried to add a constraint check by modifying your existing code:
```
"""Unit tests for all the database views"""

from django.apps import apps
from django.db import connection
from django.test import TestCase


class ViewModelAlignmentTest(TestCase):
    """Unit tests for each view. The views are automatically detected using apps.get_models()
    and filtering for models that are not managed.
    """


    def test_view_models_align_with_views(self):
        """Test that all unmanaged models representing database views have fields
        that match the columns in the database views.
        """
        unmanaged_models = [
            m
            for m in apps.get_models()
            if not m._meta.managed and not m._meta.abstract and not m._meta.proxy
        ]
        for model_class in unmanaged_models:
            with self.subTest(model=model_class.__name__):
                self.assert_view_matches_model(model_class)

    def assert_view_matches_model(self, model_class):
        """Assert that the database view associated with the given unmanaged model
        has columns and constraints that match the model's fields.
        """
        # Now expected_columns is a dictionary mapping column names to fields
        expected_columns = {field.column: field for field in model_class._meta.fields}
        view_name = model_class._meta.db_table

        with connection.cursor() as cursor:
            try:
                actual_columns = {
                    col.name: col for col in connection.introspection.get_table_description(
                        cursor, view_name
                    )
                }
            except Exception as e:
                self.fail(f"Failed to get view columns for '{view_name}': {e}")

        # Check for missing or extra columns
        missing_columns = set(expected_columns) - set(actual_columns)
        extra_columns = set(actual_columns) - set(expected_columns)

        error_message = f"Column mismatch between model '{model_class.__name__}' and view '{view_name}':\n"
        error_message += f"  Missing columns in view: {', '.join(sorted(missing_columns)) or 'None'}\n"
        error_message += f"  Extra columns in view: {', '.join(sorted(extra_columns)) or 'None'}"
        self.assertSetEqual(set(actual_columns), set(expected_columns), error_message)

        # Now, compare constraints such as `null` and `unique`
        self.check_constraints(expected_columns, actual_columns, cursor, view_name)

    def check_constraints(self, expected_columns, actual_columns, cursor, view_name):
        """Check if the null and unique constraints of the model fields match the database."""
        # Check null constraints
        for column_name, field in expected_columns.items():
            actual_column = actual_columns.get(column_name)

            # Check if column allows NULL
            if actual_column is None:
                continue  # Skip if column doesn't exist

            if field.null != actual_column.null_ok:
                self.fail(f"Field '{column_name}' null constraint mismatch: Model allows null? {field.null}, View allows null? {actual_column.null_ok}")

        # Check unique constraints
        index_info = connection.introspection.get_constraints(cursor, view_name)
        for column_name, field in expected_columns.items():
            if field.unique:
                # Check if this column has a unique constraint
                has_unique_constraint = any(
                    constraint['unique'] and column_name in constraint['columns']
                    for constraint in index_info.values()
                )
                if not has_unique_constraint:
                    self.fail(f"Field '{column_name}' is marked as unique in the model, but there is no unique constraint in the database.")
```

It seems like it works when I try to force fields to be NOT NULL or UNIQUE, but if I don't have any changes, if complains about the `deal_id` having a mismatch on our latest migration:
```
app-1  | consvc_shepherd/tests/test_database_views.py:25: 
app-1  | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
app-1  | consvc_shepherd/tests/test_database_views.py:55: in assert_view_matches_model
app-1  |     self.check_constraints(expected_columns, actual_columns, cursor, view_name)
app-1  | consvc_shepherd/tests/test_database_views.py:68: in check_constraints
app-1  |     self.fail(f"Field '{column_name}' null constraint mismatch: Model allows null? {field.null}, View allows null? {actual_column.null_ok}")
app-1  | E   AssertionError: Field 'deal_id' null constraint mismatch: Model allows null? False, View allows null? True
```

kauana (approved):
LGTM

PR #300, opened 2024-10-22
[SHEP-135] Add automatic generation of dbml file from schema on migrations

luc-lisi:
## References

[SHEP-135](https://mozilla-hub.atlassian.net/browse/SHEP-135)

## Problem Statement

We would like a way to better document the "total truth state" of our database after each migration is run. To do this, we propose dumping a `.dbml` file after each migration for better transparency on the total state of the database.

## Proposed Changes

Two new steps have been added to the `make migrate` command. Namely:
1. After `python manage.py migrate`, a `pg_dump` command is run which saves the `.sql` representation of our database's schema to a schema.sql file.
2. After the `.sql` dump completes, we convert the sql to a dbml which is more friendly for humans to read, and can be easily dropped into [dbdiagram](https://dbdiagram.io/) for easy visualization.

## Verification Steps

New dependencies required for running `sql2dbml` should be included as dev dependencies in the `package.json` file in the root directory. Verify the following:
1. `make install` correctly identifies the dev dependency and installs it
2. `make migrate` operates as normally expected _and_ a new file appears in the root directory named `schema.dmbl`
3. `make makemigrations` operates as normally expected and has the same behavior as 2. above.

[SHEP-135]: https://mozilla-hub.atlassian.net/browse/SHEP-135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -0,0 +1,8 @@
+{
```
mashalifshin, line 1: I think it might be cleaner to not do this as a dev dependency, but have this as a global install done as part of repo setup? (And documented in the `quickStart.md` of course). Then we don't need a `package.json` here, or `node_modules` at the root. And your script can go in the `/scripts` folder....

mashalifshin (commented):
Do we need to check in the schema.sql? I'm thinking we don't need both, and the DBML is the concise, readable one. So perhaps we can throw away this interim schema.sql after generating our DBML? Or do you think it has value to check in?

mashalifshin:
Oh I missed Question 3 at first, sorry! Looks like the `/schema` folder is for json schema? ... Rails puts `schema.rb` at the root, so I'd be good with the root. Also good with using the `/schema` folder for this as well, as it's clear enough. Not too opinionated here

luc-lisi:
> Do we need to check in the schema.sql? I'm thinking we don't need both, and the DBML is the concise, readable one. So perhaps we can throw away this interim schema.sql after generating our DBML? Or do you think it has value to check in?

Yeah great point! I think if we are set on using `dbdiagram` as the solution then I'd prefer throw away the `.sql` file just to keep clutter to a minimum. I'll add that cleanup to the script

luc-lisi:
> Oh I missed Question 3 at first, sorry! Looks like the `/schema` folder is for json schema? ... Rails puts `schema.rb` at the root, so I'd be good with the root. Also good with using the `/schema` folder for this as well, as it's clear enough. Not too opinionated here

Yeah it didn't seem like it belonged in `schema/` based on what was already there, but I wasn't sure. If rails does the root directory that's good enough for me though!

package.json
luc-lisi: Okay yeah glad you brought this up! I went back and forth here a ton. The thing I don't love about the global install is that weird things can happen when you need to switch node versions for other projects. Ideally everyone is using something like `nvm` to manage their different node versions, but if you do it appears you have to install the CLI over all the versions _or_ set some hacky path stuff to make sure the command is always available. This was really the only method I found to make it as "plug and play" as possible. I do agree though that it is super silly to just have this single package.json for our migration process. 

I ultimately don't have a crazy strong opinion here happy either way but wanted to share the cons I saw for the global install for context as it swayed me into the `package.json` approach. 

I'll also do some deeper diving to see if there is a better solution than the things I found so far for dealing with global installs and node version management :)

luc-lisi: (Also worth noting I'm not super familiar with managing global packages in npm anyway, so if I'm missing an easy solution or making it more complicated than it needs to be here definitely let me know haha)

mashalifshin: Hmmm yes good thinking. We do absolutely prescibe using a node version manager for the `ad-ops-dashboard`, but that isn't really part of the python/django side or it's install steps ... I guess if we are thinking of this whole thing as a single app (with a python part and a node part), and from that perspective, we can assume the developer has set up nvm, installed a node, and has a version of npm installed. So I think a global install step would mayyyyybe be okay to add to `quickStart.md`...

And then yes when we upgrade node versions, we'd need to reinstall the CLI tool, which isn't great. But upgrading node versions is a moderately big deal anyway, since we'd also be updating the way our deployed images are getting built ... (that isn't implemented yet though since we're not building the app for deploy yet)

Gonna throw a couple more ideas your way quickly before I disappear into a many-hour training
- Potentially hacky, but could we install this as part of the ad-ops-dashboard install process, which already has a package.json, perhaps as a dev dependency, and then use it from here?
- Could we hook into the existing `Makefile` commands for installing the project and get the dbdiagram that way? We do programatically require NPM there iirc, so I suppose we're already depending on someone to have set up node and npm before running those make commands, so from there I think can just add a command like `npm install -g dbdiagramcli` (or whatever it's called) and be all set ...

Having written it all out, I'm liking putting it in the `Makefile` best, but again, going quickly here so I leave it to your judgement and exploration!

luc-lisi: Added it to the make file! It will now check to see if there already exists a global install for the CLI tool and, if not found, will install it for the user when `make migrate` is called. I've also added it to the readme!

Makefile
```diff
@@ -111,15 +111,21 @@ local-test: local-test-django local-test-react   ##  Run all tests when developi
 makemigrations-empty: ##  Create an empty migrations file for manual migrations
 	docker exec -it consvc-shepherd-app-1 python manage.py makemigrations --empty consvc_shepherd
 
-migrate: ##  Run migrate on the docker container
+migrate: ##  Run migrate on the docker container. This will run Django migrations and create .sql, .dbml, and visual files
+	npm list -g @dbml/cli || npm install -g @dbml/cli
 	docker exec -it consvc-shepherd-app-1 python manage.py migrate
+	docker exec -it consvc-shepherd-db-1 pg_dump -U postgres -s -F p -E UTF-8  postgres > schema.sql
+	sql2dbml schema.sql --postgres -o schema.dbml
+	@echo "Cleaning .sql file as it's no longer needed..."
```
mashalifshin, line 119: Total nit but I don't think you need to log this here

docs/quickStart.md
```diff
@@ -55,13 +58,20 @@ See the [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) documentat
 1. Install your dependencies:
 
     ```shell
-    $ pip install poetry
-    $ poetry install
+    pip install poetry
+    poetry install
+    ```
+
+    We also require a global dependency to be installed now in order to make migration diagrams automagically in `dbml format.`
```
mashalifshin, line 65: Further down in these instructions, we tell folks to install node and npm. Might want to mention that here, that they can jump down to those and install node if they don't have it already. Or, we could potentially have the installation of the global DB diagramming stuff happen after the node installation steps down below. 

I think it would also be nice to link to the DBML docs here so folks can reference if they aren't familiar.

docs/quickStart.md
```diff
@@ -137,26 +151,28 @@ It is recommended to use a node version manager to allow for multiple `node`s on
...
-    $ cd ..
-    $ docker-compose up --build
+    cd ..
+    docker-compose up --build
     ```
 
     (You can also run the `ad-ops-dashboard` without the API via `npm run dev` from `./ad-ops-dashboard`)
 
-7. Visit the home page at http://0.0.0.0:5173/
+7. Visit the home page at <http://0.0.0.0:5173/>
```
mashalifshin, line 178: Thank you for cleaning up this doc! And I especially  appreciate your attention to the details like making this a link!

docs/quickStart.md
mashalifshin: Oh, we could also mention here that you can copy-paste that schema.dbml file into the dbdiagram.io UI to get a diagram (that will change when we do the followup to get the png output right into the repo, but for now seems helpful to mention)

Do we need to create a ticket for turning this into a diagram? IIRC that's blocked on getting a license, but still would be good to capture in Jira

mashalifshin (approved):
This works like a charm! I also really like how clean the code ended up here. Thank you for your attention to the docs here as well. 

Verified the script on my local and got the dmbl output, and it works when pasted into the dbdiagram.io UI :shipit:

Makefile
luc-lisi: Removed!

docs/quickStart.md
luc-lisi: Ah yeah more documentation is always a good call!

I've updated those docs with all the points you mentioned 😄 

Also created a ticket for next steps once we have a license for dbdiagram here! https://mozilla-hub.atlassian.net/browse/SHEP-163

mashalifshin: The updates to the docs are really nice, thank you! 🔥

package.json
mashalifshin: I see your `npm list -g @dbml/cli || npm install -g @dbml/cli`, that's awesome!

PR #302, opened 2024-10-22
[SHEP-36] Adding additional flexibility to local logging

luc-lisi:
## References

[SHEP-36](https://mozilla-hub.atlassian.net/browse/SHEP-36)

## Problem Statement

After speaking with @mashalifshin it sounds like we wanted to expand on this ticket to include an entire custom local logging formatted. This will allow us to not only add human readable timestamps, but also allow for us, in the future, to have more verbose, "local specific" logging that can help with debugging.

## Proposed Changes

We've added a new formatter called `localdev` to `settings.py` file. This is enabled when the env variable `CUSTOM_LOCAL_LOGGER_ENABLED` is set to true and disabled by default or when explicitly set to false. This new logging formatter will display most of the fields similar to the JSON logger with some small changes.

1. The log level has been moved to a header and in human readable format (rather than a number value)
2. The timestamp has been changed to be in human-readable format rather than UNIX ns. 

We've also gone ahead and removed the `APP_ENV` variable as this new flag is much clearer as to the intent.

## Additional Changes

I noticed a bug in running the sync_boostr script. It seems like if the rate limit has been hit, the app will fail on startup due to the fact the `log` object is initialized _after_ the first call to the APIs are made. As a result, I've moved the declaration of the logger to the first line of the init. Seeing as this is still related to logging I've included it in this ticket but happy to make it a new ticket if people prefer. 

## Verification Steps

1. Pull down this branch and run the local script as you normally would. You should see no changes to the standard logging behavior of the application.
2. Add a new `.env` variable: `CUSTOM_LOCAL_LOGGER_ENABLED=true`.
4. Run the script normally and verify the logs now appear to be in the form `INFO {Timestamp: 2024-10-22 15:40:55,256, Type: sync_boostr_data, Logger: shepherd, Message: 'Upserted products and budgets for deal: 435629', Pid: 42}`

## Visuals

Example of the logging format with this new flag set to true.
Screenshot 2024-10-22 at 8 41 07 AM

[SHEP-36]: https://mozilla-hub.atlassian.net/browse/SHEP-36?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
 DEBUG=true
-APP_ENV=dev
+CUSTOM_LOCAL_LOGGER_ENABLED=false
 SHEPHERD_ENV=INFO
 SECRET_KEY=keyboard-mash
 DB_NAME=postgres
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ doc: ##  Generate docs via mdBook
 doc-preview: doc  ##  Preview Merino docs via the default browser
 	mdbook serve --open
 
-dev: $(INSTALL_STAMP)  ## Run shepherd locally and show human readable timestamps.  
+dev: $(INSTALL_STAMP)  ## Run shepherd locally and show human readable timestamps.
 	docker compose up -d && docker-compose logs -f -t
 
 local-test-django: $(INSTALL_STAMP) # Run shepherd Django app tests locally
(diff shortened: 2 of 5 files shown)


mashalifshin:
Seems like @kauana is the reviewer here, but I just wanted to say I really love how clean this approach turned out, and how we have a place to add more custom logging in dev as a result. Also appreciate the care re. setting the default on that env var to false!

kauana (approved):
I ran the scripts twice and confirmed that we only switch to local logs if we have `CUSTOM_LOCAL_LOGGER_ENABLED=true` set. 

Thanks Luc! The logs look great 🎉

PR #307, opened 2024-10-23
[SHEP-153] Removing redudant env variable definition for local dev

luc-lisi:
## References

[SHEP-153](https://mozilla-hub.atlassian.net/browse/SHEP-153)

## Problem Statement

This env variable`DJANGO_SETTINGS_MODULE` is a redundant definition and can cause things to pass locally that might fail in prod/stage. 

## Proposed Changes

Removing this env variable.

## Verification Steps

Verify application works as normally expected, tests pass, and scripts run.

[SHEP-153]: https://mozilla-hub.atlassian.net/browse/SHEP-153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kauana:
@luc-lisi if we remove `DJANGO_SETTINGS_MODULE` from the `.env` file we also need to set it when running the django tests locally with `local-test-django` / `local-test` otherwise all tests fail. Note that this is already done for the [tests running in CI ](https://github.com/mozilla-services/consvc-shepherd/blob/48a13678decfde5e26664dd7196be00a0410202c/Makefile#L84)(and why they pass)

luc-lisi:
@kauana This is such a good catch! Thank you!

luc-lisi:
@kauana I've gone ahead and added it to the docker compose file for the tests. We should not however need it in our main `docker-compose.yml` file as manage.py scripts are the first to run and the os.environ variable is set there. Thanks again for catching this! It's good we are cleaning up all these caveats where bugs could hide 😄 

Let me know if you'd like any other modifications!

kauana (approved):
Thanks! I've confirmed that our tests work now without `DJANGO_SETTINGS_MODULE` set in `.env` 🚢

PR #312, opened 2024-10-29
[DRAFT] [SHEP-161] Update linting rules

luc-lisi:
## References

[SHEP-161](https://mozilla-hub.atlassian.net/browse/SHEP-161)

## Problem Statement

We would like to enforce a slightly stricter style within the python repo as well as move towards `ruff` as the sole linter for our project.


Makefile
```diff
@@ -65,9 +65,20 @@ eslint-fix: $(INSTALL_STAMP)  ##  Format code with eslint
 	cd ad-ops-dashboard && npm run lint:fix
 
 # Temporarily disable eslint while we fix the current issues in the ad-ops-dashboard
-lint: $(INSTALL_STAMP) isort black flake8 bandit pydocstyle mypy eslint ##  Run various linters
+lint: $(INSTALL_STAMP)
```
mashalifshin, line 68: Love that we're gonna run 2 linters'n'formatters, and not all the ones in the known python universe!

docs/style.md
```diff
@@ -0,0 +1,77 @@
+# Sheherd Style Guide
```
mashalifshin, line 1: I think it's great we are adding a `style.md` file!

This doc seems like it's written from the POV of advocating for/convincing folks that it's a good idea to lint, type check, etc. I think it's fine to just say "this is how it is" and remove the places where you are justifying the choice for these industry standard code quality practices. This is totally a gray area and up to you what feels relevant, but for an example from my end, the context re. "why pyright" feels informative for a dev, whereas the context in the "static type checking" section seems like its preaching to the choir.

mashalifshin (commented):
Super on board with everything proposed here, specifically
- big linting pass PR over the whole repo with new rules
- list of rules we want to adopt as time/priority allows
- unifying on a couple lint/format tools 
- style.md 🙌

docs/style.md
luc-lisi: Ah yes! Great call out. As I was brain dumping in this readme I totally was trying to include motivation for all of this, but you're absolutely right that it's excessive for the intended audience. I love how _stoked on style_™ this team is 🥳 

I'll edit down this readme to be better aimed at where our team is at and thank you again for the comment! Absolutely do not want to come off as preachy here.

Makefile
luc-lisi: Yes!! And what's crazy is that the performance for running `ruff` with all the rules is like 10x faster than most of these tools individually 🚤

consvc_shepherd/.models.py.swp
kauana, line 1: I see a vim temp file 👀 . Feel free to add it to our `.gitignore` or alternatively you can configure vim to save these files in a separate folder.

Makefile
```diff
@@ -65,9 +65,20 @@ eslint-fix: $(INSTALL_STAMP)  ##  Format code with eslint
...
+	@echo "Running Mypy..."
+	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
+	@echo "Running Ruff..."
+	$(POETRY) run ruff check
+	# eslint
 
-lint-fix: $(INSTALL_STAMP) isort-fix black-fix flake8 bandit pydocstyle mypy eslint-fix ##  Run various linters and fix errors to pass CircleCi checks
+lint-fix: $(INSTALL_STAMP)
```
kauana, line 75: I haven't run the linter yet, but I just wanted to point out that Nana and I had some discussions earlier about changing ruff to replace `bandit`. I'm curious to know if you found that ruff is similar or better than bandit at spotting some securities issues in our code

luc-lisi: Hi @kauana! Great question!

We do have `ruff` configured with [Bandit rules](https://docs.astral.sh/ruff/rules/#flake8-bandit-s).

The example snippet you quoted will return the following linting errors using this updated ruleset:

```
test.py:1:1: D100 Missing docstring in public module
test.py:7:5: D103 Missing docstring in public function
test.py:9:9: S506 Probable use of unsafe `yaml.load`. Allows instantiation of arbitrary objects. Consider `yaml.safe_load`.
test.py:18:5: D103 Missing docstring in public function
test.py:20:5: F841 Local variable `j` is assigned to but never used
test.py:23:24: S506 Probable use of unsafe loader `Loader` with `yaml.load`. Allows instantiation of arbitrary objects. Consider `yaml.safe_load`.
```
If there is anything missing though, definitely let me know! Running Bandit directly it looks like these two security rule errors we get in ruff (S506) should have parity with the Bandit rule B506 😄

consvc_shepherd/.models.py.swp
luc-lisi: Good catch! Fixed :)

kauana:
@luc-lisi do you think we should also remove the linter dependencies (flake8, bandit, etc) from [poetry.lock](https://github.com/mozilla-services/consvc-shepherd/blob/main/poetry.lock)?

consvc_shepherd/admin.py
```diff
@@ -464,19 +446,12 @@ def get_queryset(self, request):
...
-            )
-            or None
-        )
+        return ", ".join(str(flight.kevel_flight_id) for flight in obj.flights.all() if flight.kevel_flight_id) or None
 
 
 @admin.register(CampaignSummary)
 class CampaignSummaryAdmin(admin.ModelAdmin):
-    """Admin model for showing the Campaign overview"""
+    """Admin model for showing the Boostr deals revenue overview."""
```
kauana, line 454: hmmm we should keep `Campaign overview` here instead of `Boostr deals revenue overview` because:
1. Campaign in this context refers to a "Shepherd" campaign. In fact, Boostr doesn’t really have the concept of a campaign. In Boostr a campaign is actually represented as a media line item. 
2. View includes more than just revenue information, such as click and impression data.

pyproject.toml
```diff
@@ -55,21 +57,15 @@ types-mock = "^5.1.0"
...
-
 [tool.mypy]
 python_version = "3.11"
+#disallow_untyped_defs = true
 disallow_untyped_calls = true
 follow_imports = "normal"
 ignore_missing_imports = true
+#disallow_any_unimported = true
+#no_implicit_optional = true
+#check_untyped_defs = true
```
kauana, line 68: I think we want to clean up these mypy settings?

pyproject.toml
```diff
@@ -47,6 +48,7 @@ pytest-django = "^4.5.2"
 pytest-cov = "^4.0.0"
 selenium = "^4.8.0"
 mypy = "^1.2.0"
+pyright = "^1.1.386"
```
kauana, line 51: I think we should wait until the meeting next week to discuss if we should add pyright, right? :P

docs/style.md
```diff
@@ -0,0 +1,83 @@
...
+
+
+#### Why Pyright?
+
+Mypy has been around forever, why would we use Pyright? It's a great question. The "tl;dr" version is that Mypy is only really an implementation of PEP 484. This is great, but it is not exhaustive and many type checking behaviors are left, intentionally, undefined. Pyright instead references the official [Python typing specification](https://typing.readthedocs.io/en/latest/spec/index.html). As a result, it simply has more features than Mypy, and is often more accurate. You can see [this page](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html) for up-to-date benchmarks of mypy and pyright vs the Python type system specs.
+
+It's also faster, has great plugin support for VSCode (as pyright is also managed by Microsoft), and is stricter "out of the box" when compared to mypy. In a modern repo, it's a strong choice.
+
+If you're still not convinced though, check out [this page](https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md) for a full comparison between the two. Admittedly this comparison is bias (as it's written by the pyright maintainers), but the content is still accurate.
```
kauana, line 83: do we want this Pyright section added as part of this PR, or should we include it only if and when we decide to switch to Pyright?

openidc/middleware.py
```diff
@@ -23,8 +23,8 @@ def validate_iap_jwt(request):
             certs_url="https://www.gstatic.com/iap/verify/public_key",
         )
         return decoded_jwt["email"]
-    except Exception as e:
-        logger.error(f"IAP JWT validation error: {e}")
+    except Exception:
+        logger.exception("IAP JWT validation error")
```
kauana, line 27: I've seen a couple instances of exceptions that haven't been raised. I'm wondering if we've decided to exclude these on purpose or if we should indeed raise all exceptions?

consvc_shepherd/management/commands/sync_boostr_data.py
```diff
@@ -378,44 +351,36 @@ def upsert_deal_products(self, deal: BoostrDeal) -> list:
         return deal_products
 
     @classmethod
-    def update_sync_status(self, status: str, synced_on: datetime, message: str):
-        """Update the BoostrSyncStatus table given the status and the message"""
+    def update_sync_status(cls, status: str, synced_on: datetime, message: str):
```
kauana, line 354: I'm not seeing the definition of `cls` here so I think this is a typo?

consvc_shepherd/api/urls.py
```diff
@@ -1,4 +1,4 @@
-"""aconsvc_shepherd API URL Configuration
+"""aconsvc_shepherd API URL Configuration.
```
kauana, line 1: typo: `aconsvc_shepherd` > "consvc_shepherd"

kauana:
Tests done:
- ran all the tests with `make local-test` and confirmed they passed
- ran sync boostr data and checked that its working
- ran sync bq data and checked that its working
- looked through all the uis that have data (except Allocation/Settings Snapshot because I don't really know how to properly test them but we can double check with Glenda)

Makefile
```diff
@@ -65,9 +65,20 @@ eslint-fix: $(INSTALL_STAMP)  ##  Format code with eslint
 	cd ad-ops-dashboard && npm run lint:fix
 
 # Temporarily disable eslint while we fix the current issues in the ad-ops-dashboard
-lint: $(INSTALL_STAMP) isort black flake8 bandit pydocstyle mypy eslint ##  Run various linters
+lint: $(INSTALL_STAMP)
+	@echo "Running Mypy..."
+	$(POETRY) run mypy $(APP_DIRS) --config-file="pyproject.toml"
+	@echo "Running Ruff..."
+	$(POETRY) run ruff check
+	# eslint
```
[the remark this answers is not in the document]
luc-lisi: Hmmm, I get the error when I run locally that `eslint: no such file or directory`

In the comment above this make file it says `# Temporarily disable eslint while we fix the current issues in the ad-ops-dashboard`. So figured it might just not be something implemented yet. Do we know if these have been resolved?

openidc/middleware.py
luc-lisi: Yeah if I'm being honest I'm not entirely sure why the author of this particular section of code is wanting to just log this rather than re-raise it. I suppose maybe we just want transparency into this error rather than causing a potentially breaking Exception raise? But didn't want to mess with it in case it had unseen consequences. 

The change here was just because logger.exception is just meant to be used in-place of logger.error when inside an try/except statement as `logger.exception` will automatically include information about the exception raised. So the new rules just recognize `logger.error(f"{e}")` as an anti-pattern. (https://docs.astral.sh/ruff/rules/error-instead-of-exception/)

consvc_shepherd/api/urls.py
luc-lisi: Good eye! Fixed!

pyproject.toml
luc-lisi: Agreed! Done

ruff.toml
```diff
@@ -5,47 +5,75 @@ indent-width = 4
...
+  "venv",
+  "migrations",
+  "manage.py",
+  "*test_*.py",
 ]
 
 [lint]
-select = ["E","W","S","I","B","D202","D212"]
-ignore = ["D105","D107","D203", "D205", "D400","S101", "S104"]
+# select = ["E","W","S","I","B","D202","D212"]
```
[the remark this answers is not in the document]
luc-lisi: Ah yes! This was left in by accident. Removed!

ruff.toml
```diff
@@ -5,47 +5,75 @@ indent-width = 4
...
+]
+
 fixable = ["ALL"]
 
 [lint.isort]
 order-by-type = true
 
 [lint.pydocstyle]
-convention = "pep257"
+convention = "pep257" # Do we want to revist this?
```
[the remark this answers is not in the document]
luc-lisi: Ah yes great call! Left that comment in to remind me to bring it up. Comment removed!

consvc_shepherd/admin.py
luc-lisi: Oh great catch! I have no idea how this change happened. Something weird might have happened when I rebased? Either way should be fixed!

luc-lisi:
> @luc-lisi do you think we should also remove the linter dependencies (flake8, bandit, etc) from the Makefile and from [poetry.lock](https://github.com/mozilla-services/consvc-shepherd/blob/main/poetry.lock) or should we do this in a later PR?

Oh that's a great call! I'll remove them in this PR 😄

docs/style.md
luc-lisi: Oh that's a good point. Let me get that into another branch and remove these dependencies for now 😄

pyproject.toml
luc-lisi: Haha great point! I'll remove that from this PR 😄

Makefile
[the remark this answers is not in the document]
kauana: Wait I'm confused 🤔 I'm [looking at main](https://github.com/mozilla-services/consvc-shepherd/blob/ecd1219bbc328faa36efc2507fb10130f2b74a11/Makefile#L70) we **are** running it, but maybe we forgot to remove the comment?

consvc_shepherd/management/commands/sync_bq_data.py
```diff
@@ -110,28 +116,30 @@ def query_bq(self) -> pandas.DataFrame:
...
-            self.log.error(f"An error occurred while querying BigQuery: {e}")
-            raise  # Re-raise the exception to propagate it further
+            self.log.exception("An error occurred while querying BigQuery")
+            raise SyncBQDataError(e) from e  # Re-raise the exception to propagate it further
+        else:
+            return df
+
+    def _error_if_empty_result(self, result: Union[RowIterator, _EmptyRowIterator]):
+        if result.total_rows == 0:
+            raise NoDataReturnedError(self.date)
```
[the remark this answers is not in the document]
luc-lisi: Oooh good point! Yeah we should 100% discuss this then. This convention is from the [official python docs](https://docs.python.org/3/tutorial/classes.html#private-variables) so I'd definitely advocate for folks to use it where appropriate as the standard library follows this convention as well. But happy to remove it for now if this isn't something the whole team agrees on!

Makefile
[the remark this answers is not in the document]
luc-lisi: Oh weird 👀 

Does it work locally when you run `make lint` with eslint uncommented?

openidc/middleware.py
mashalifshin: this is totally at a glance, and applies only for this specific situation, but it seems normal/intended to not throw for a jwt validation error. many of the auth libs I've worked with throw for this case, but it's not exactly an error in the application programming -- it just means the jwt didn't pass validation which could happen in the normal course of auth flows such as an expired token, or a truly invalid visit to the url ...

luc-lisi: Ah that makes sense. In that case probably best to leave it as is? Either way I worry any logic changes are probably outside the scope of this PR anyway, but let me know if y'all disagree!

kauana: Sounds good! Thanks folks

kauana (approved):
Thanks for quickly addressing all comments! LGTM 🎉

PR #315, opened 2024-10-30
[SHEP 159] Add Revenue Overview view

kauana:
## References

- [SHEP-159](https://mozilla-hub.atlassian.net/browse/SHEP-159)

## Problem Statement

As part of our [MVP](https://docs.google.com/document/d/1V_oweGAsIjAWkWpfAGas5tezVXydJaDPpkRwqIuhyas/edit?tab=t.0#heading=h.33nf0vnev3p2), Ad Ops needs to be able to see a summary of product revenue. This PR is the initial implementation of this feature on the Django side only. The React work will come in a later PR.

## Proposed Changes

- **Add a new admin page for `BoostrDealProduct`**: This page was previously not available as there was no need to relate data between Deals and Products. However, it proved useful for comparing product revenue data across multiple deals, so this page has been added to support future revenue information.
  - Also, add Product and Year filtering to the BoostrDealProduct UI.

- **Create a new view, "Revenue Overview"**: This view will display product revenue as outlined in the MVP:
  - `product`: Boostr product name
  - `projected_revenue`: Sum of all deal weighted amounts for a given product

- **React lint warning fix**: Fixes `React Hook useEffect has a missing dependency: 'debouncedOnFilterChange'` warning displayed when we ran `make lint-fix`.

- **Add `schema.sql` file**: Luc's initial PR (#300) didn't include the `schema.sql` file that gets generated when running `make migrate` because we thought it would be redundant. However, we’ve learned that `schema.dbml` doesn’t display views, whereas `schema.sql` does, so we've decided to add it.

- **Add tests**:
  - Add unit tests for `RevenueOverview` view, `BoostrDealProduct` tests.

- **Model and Admin cleanup**:
  - Remove redundant Meta verbose classes, add missing/updating docstrings, and make formatting changes for consistency.

## Verification Steps

1. Testing this involves checking that the view is calculating the revenue correctly. You can test this in two ways:
   1. With mock data by inserting values into the database. Here is an example:

      ```sql
      INSERT INTO consvc_shepherd_boostrproduct (id, boostr_id, full_name, campaign_type, country, created_on, updated_on)
      VALUES
          (100, 1001, 'Product Alpha', 'CPM', 'US', NOW(), NOW()),
          (200, 1002, 'Product Beta', 'CPC', 'CA', NOW(), NOW());

      INSERT INTO consvc_shepherd_boostrdeal (id, boostr_id, name, advertiser, currency, amount, stage, sales_representatives, start_date, end_date, created_on, updated_on)
      VALUES
          (10000, 2001, 'Deal One', 'Advertiser A', '$', 500, 'Closed Won', 'John Doe', '2024-01-01', '2024-12-31', NOW(), NOW()),
          (20001, 2002, 'Deal Two', 'Advertiser B', '$', 1000, 'Verbal', 'Jane Doe', '2024-02-01', '2024-12-31', NOW(), NOW()), 
          (20002, 2003, 'Deal Three', 'Advertiser A', '$', 800, 'Renewal', 'Alice Smith', '2024-03-01', '2024-12-31', NOW(), NOW()), 
          (20003, 2004, 'Deal Four', 'Advertiser C', '$', 100, 'Closed Won', 'Bob Johnson', '2024-04-01', '2024-12-31', NOW(), NOW()), 
          (20004, 2005, 'Deal Five', 'Advertiser D', '$', 2200, 'Renewal', 'Chris Lee', '2024-05-01', '2024-12-31', NOW(), NOW()), 
          (20005, 2006, 'Deal Six', 'Advertiser B', '$', 2000, 'Renewal', 'Dana White', '2024-06-01', '2024-12-31', NOW(), NOW()), 
          (20006, 2007, 'Deal Seven', 'Advertiser C', '$', 1500, 'Verbal', 'Emily Green', '2024-07-01', '2024-12-31', NOW(), NOW());

      INSERT INTO consvc_shepherd_boostrdealproduct 
          (id, boostr_deal_id, boostr_product_id, budget, month)
      VALUES
          (101, 10000, 100, 500, '2024-01'),
          (102, 20001, 100, 1000, '2024-02'),
          (103, 20002, 100, 800, '2024-03'),
          (104, 20003, 200, 100, '2024-04'),
          (105, 20004, 200, 2200, '2024-05'),
          (106, 20005, 200, 2000, '2024-06'),
          (107, 20006, 200, 1500, '2024-07');
      ```

      The example above should show:

       image

    Explanation:
        * Product A
             - `projected_revenue` = (500 * 1.0) + (1000 * 0.9) + (800 * 0.75) = 2000
        * Product B
             - `projected_revenue` = 100 + (2200 * 0.75) + (2000 * 0.75) + (1500 * 0.90) = 4600

   2. With Boostr data. Most products have multiple deals, but you can filter by the `Pocket.com Combined` product and we shouldn't have too many entries. We can calculate the `projected_revenue` accordingly to ensure that the view is correct.

## Visuals

Pull the changes and take a look at the [local Revenue Overview page](http://0.0.0.0:7001/admin/consvc_shepherd/revenueoverview/).

[SHEP-159]: https://mozilla-hub.atlassian.net/browse/SHEP-159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


schema.sql
```diff
@@ -0,0 +1,1523 @@
...
+);
+
+
+ALTER TABLE public.django_session OWNER TO postgres;
+
+--
+-- Name: revenue_overview_view; Type: VIEW; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.revenue_overview_view AS
```
luc-lisi, line 794: Could you re-run the `make migrate` script? It looks like when we manually changed the CREATE VIEW -> CREATE TABLE we didn't revert that change!

kauana: oops! good catch

consvc_shepherd/models.py
```diff
@@ -569,20 +573,52 @@ def live(self):
...
+    product: CharField = models.CharField()
+    projected_revenue: DecimalField = models.DecimalField(
+        max_digits=10, decimal_places=2
+    )
+    target_revenue: DecimalField = models.DecimalField(max_digits=10, decimal_places=2)
+
+    @property
+    def revenue_delta(self):
+        """Difference between target_revenue and projected_revenue"""
+        if self.projected_revenue and self.target_revenue:
```
luc-lisi, line 610: I'm not 100% sure how these `DecimalField` types are represented here, but are we checking here for `None` values here, or `None` and `0` values? Python will consider values of `0` to be falsey so intended business logic isn't super clear from just reading this. For example, if `target_revenue = 200` and `projected_revenue = 0`. As it stands it _might_ return `None` or it _might_ return `200` depending on how Python handles the truth value for a x: DecimalField = 0.

All this to say, do we know the logic we are intending here for this edge-case and is there a way we can make this conditional more explicit? Unless this is a totally ridiculous edge-case based on how the data would actually look haha. Just wasn't sure!

consvc_shepherd/tests/test_models.py
```diff
@@ -120,6 +124,26 @@ def test_campaign_summary_fields(self):
...
+
+    def setUp(self):
+        """Set up test data for RevenueOverview model."""
+        self.boostrproduct = BoostrProductFactory(boostr_id=1, full_name="Product1")
+        self.boostrdeal = BoostrDealFactory(stage="Verbal")
+        self.boostrdealproduct = BoostrDealProductFactory(
+            boostr_deal=self.boostrdeal, boostr_product=self.boostrproduct, budget=1000
+        )
+
+    def test_revenue_overview_fields(self):
```
luc-lisi, line 138: Following up on my previous comment, it might be good to have a test case here when target_revenue or projected_revenue is `0`! Just to make sure the conditional logic in `revenue_delta` is what we want it to be 😄 .

consvc_shepherd/tests/test_admin.py
```diff
@@ -385,6 +385,72 @@ def test_delete_allocation(self) -> None:
...
+            budget=7000,
+            month="2024-02",
+        )
+
+    def test_year_filter(self):
+        """Test filtering BoostrDealProduct by year"""
+        response = self.client.get(
+            reverse("admin:consvc_shepherd_boostrdealproduct_changelist")
+            + "?year=2023",  # filter by year 2023
+            **{"settings.OPENIDC_HEADER": "dev@example.com"},
```
luc-lisi, line 436: I'm noticing that all other tests look like this, but is this correct? It seems like the 3rd positional arg for `client.get` is `follow: bool`. Is this meant to be the `extra` kwargs? If so it might be worth explicitly changing the call to be:

```python
response = self.client.get(
            reverse("admin:consvc_shepherd_boostrdealproduct_changelist")
            + "?year=2023",  # filter by year 2023
            extras={"settings.OPENIDC_HEADER": "dev@example.com"},
        )
```
Unless this is some super secret python shorthand I don't know about!

consvc_shepherd/tests/test_admin.py
```diff
@@ -385,6 +385,72 @@ def test_delete_allocation(self) -> None:
...
+        )
+
+        print(response.content.decode())
```
luc-lisi, line 439: Possible left over print statement from local testing!

luc-lisi (commented):
Great PR! Tested locally and everything runs as expected 🎉 

Left a few quick comments but definitely not a Django expert, so if I'm misguided in any of them definitely let me know! 😄

consvc_shepherd/tests/test_admin.py
kauana: 100% sneaky print statements 😜

consvc_shepherd/tests/test_admin.py
luc-lisi: Looking into this more it seems like python actually is able to be clever about this and know that the unpack in a function call is explicitly for kwargs. So I think seeing as this is how all the other `client` calls are being made, we can probably ignore this for now to keep everything consistent.

Maybe this is something we can look into though later as we add stricter linting rules to the repo 😄 

So please ignore the above comment for now!

mashalifshin:
> target_revenue: Sum of all deal amounts for a given product

I'm jumping into the comments on this PR from our thread on the design doc, hope that's okay. 

I don't think this is the correct definition of "Target revenue". "Target revenue" is a set of yearly goals set for the Sales organization, and those are defined in [yet another spreadsheet](https://docs.google.com/spreadsheets/d/1_M3FXIkQT947TGmaKw_Mhy1OjUN9h4AP-WIJams3tSo/edit?gid=1609652617#gid=1609652617). 

So I don't think we want to show this column for this iteration of the Revenue Overview. That is totally okay and intended, since this version of the overview is focused on surfacing data from Boostr (M1) and BQ (M2), and we don't need to worry about showing data from any other sources in this iteration. 

In addition to that, it sounds like Laura doesn't even need "Target Revenue" or "Revenue Delta" columns in the Revenue Overview at all, as those columns are more relevant to how sales are progressing as opposed to how campaigns are delivering against their targets. We should definitely confirm that with her, but I think for the purpose of this PR it's safe to drop those columns from the overview.

Lmk if that is making sense or if it would be helpful to hop on a call...

kauana:
> > target_revenue: Sum of all deal amounts for a given product
> 
> I'm jumping into the comments on this PR from our thread on the design doc, hope that's okay.
> 
> I don't think this is the correct definition of "Target revenue". "Target revenue" is a set of yearly goals set for the Sales organization, and those are defined in [yet another spreadsheet](https://docs.google.com/spreadsheets/d/1_M3FXIkQT947TGmaKw_Mhy1OjUN9h4AP-WIJams3tSo/edit?gid=1609652617#gid=1609652617).
> 
> So I don't think we want to show this column for this iteration of the Revenue Overview. That is totally okay and intended, since this version of the overview is focused on surfacing data from Boostr (M1) and BQ (M2), and we don't need to worry about showing data from any other sources in this iteration.
> 
> In addition to that, it sounds like Laura doesn't even need "Target Revenue" or "Revenue Delta" columns in the Revenue Overview at all, as those columns are more relevant to how sales are progressing as opposed to how campaigns are delivering against their targets. We should definitely confirm that with her, but I think for the purpose of this PR it's safe to drop those columns from the overview.
> 
> Lmk if that is making sense or if it would be helpful to hop on a call...

@mashalifshin  of course its okay! 

You are correct, `target_revenue` is not the Boostr amount, but since we don't have the real `target_revenue` values, I've decided to go ahead and use the boostr amount for now because I didn't want to block the story on this.

As for not needing the "Target Revenue" or "Revenue Delta" columns, I already had the work done by the time Laura mentioned that in the doc hence that question during standup I had asking if I should remove these two columns from this PR or keep it. I just checked again and unfortunately Laura hasn't replied to that doc again, so I'm happy to remove it now or wait to discuss this next week during the feedback meeting

kauana:
@mashalifshin @luc-lisi I went ahead and removed `target_revenue` and `revenue_delta` from this PR and added a note to our meeting notes next week to discuss this.

Luc, please take a look at the latest changes when you have the time. I changed `projected_revenue` to look more like money amount so let me know what you think:

Before:
image

After:
image

consvc_shepherd/admin.py
```diff
@@ -439,22 +502,46 @@ class CampaignSummaryAdmin(admin.ModelAdmin):
...
+        "campaign_id",
+        "flight_id",
+        "flight_name",
+    ]
+    search_help_text = "Search by product name"
+
+    @admin.display(description="Projected Revenue")
+    def formatted_projected_revenue(self, obj):
+        """Format projected revenue as a cash amount"""
+        return f"{obj.projected_revenue:,.2f}"
```
luc-lisi, line 526: This is a great addition!! I love these quality of life improvements to the admin panel 🥳 

One thing I did notice is that this kind of formatting will raise a TypeError and break the admin page if any value of`projected_revenue` is ever somehow `None`... but I _think_ that should be impossible looking at how the view is constructed.  Plus we might move away from views, which will let us better enforce the types on the fields. So totally up to you if you think we should add a None check here!

luc-lisi (approved):
Confirmed new changes work locally! Made one comment but it's non-blocking. Feel free to resolve it if you think it's a non-issue 😄 

Approved! 🚢

consvc_shepherd/admin.py
kauana: Exactly, since both BoostrDeal amount and stage can't be `null`, I don't expect `projected_revenue` to ever be `None`. However, I updated to code to display `0.00` if the `projected_revenue` is ever `None`.  I also made Revenue Overview a read only table since this is a view and we shouldn't be adding anything to use.

kauana:
> @kauana just closing the loop on the target revenue discussion. I think removing it totally works and I'm sorry I hastily said the wrong thing when we talked about it at stand up! After sleeping on this, I think the "Sum of all deal amounts for a given product" calculation sounds like it would be useful to show as well, but under a different name. Maybe we can check with Laura if such a column would be useful?

Not a problem at all! I've removed it from this PR, but if we learn that it would benefit them,  I'm add create a new ticket/address this at a later PR.

PR #316, opened 2024-10-31
[AE-636] Fix mobile tiles preview

dmueller:
## References

https://mozilla-hub.atlassian.net/browse/AE-636

## Problem Statement

In testing AMP callbacks, Glenda noticed that locally she couldn't get test mobile tiles from MARS but the preview page showed that working fine.

It turned out that there were a couple problems:
* preview wasn't updated with mobile placement names
* the UAPI requests are not specifying the user agent, so MARS was always determining them to be a desktop client

This made the tiles returned always come from the desktop AMP endpoint, and then they were rendered with the styles appropriate for mobile.

## Proposed Changes
This adds the mobile tile placement names and uses the FormFactors to drive that condition. It also adds the User-Agent header to UAPI requests.

## Verification Steps
Ran locally and verified that dev UAPI and prod UAPI both returned ads for desktop and mobile. This didn't really confirm the code was correct since the preview page was already showing ads for those; but it did show some small differences.

AMP returns 2 tiles for prod mobile, so even if we request 3 we only get 2 back. This difference is visible now.
AMP didn't have test tiles set up for mobile when I started this PR -- the thing that made us realize this gap. While it was a WIP there was some time where dev mobile tiles were an empty and the desktop dev tiles were not. Glenda has reached out to AMP to ask about this, and as of this morning it looks like there are some tiles in both now.


consvc_shepherd/preview.py
```diff
@@ -137,7 +137,7 @@ class Ads:
     Environment(
         code="dev",
         name="Dev",
-        mars_url="https://mars.stage.ads.nonprod.webservices.mozgcp.net",
+        mars_url="https://ads.allizom.org",
```
dmueller, line 140: the url adjustments here were not a problem, we just have the short urls now so might as well get everything using them

consvc_shepherd/tests/test_preview.py
```diff
@@ -382,13 +392,17 @@ def mock_unified_data(self, *args, **kwargs):
...
             ],
         }
-        return mock.Mock(status_code=200, json=lambda: response)
 
-    def test_get_unified(self):
-        """Test the Retrieval of Ads from MARS Unified API."""
+        def mock_data(self, *args, **kwargs):
+            return mock.Mock(status_code=200, json=lambda: response)
+
+        return mock_data
```
dmueller, line 399: I'm not sure if there is a better way to have the mock response have a more dynamic configuration; but in order to have it change based on form factor I still needed to have a function returned for the argument in mock.patch

mashalifshin: Yeah I'm a bit confused by how that mock function is currently set up, seems like if we're using it to `mock.patch` the `requests.post` function, then it should accept the same args, so that those could be used to decide what different mock data to return ... But I think what you did is great given the existing setup

mashalifshin (approved):
Great detective work @dmueller and @gleonard-m for catching this issue! 

I feel like this isn't the first time we've had a situation where preview wasn't correctly calling UAPI and was obscuring important info, and I'm wondering if we can catch this more proactively somehow ...

mashalifshin:
I tweaked the format for your PR title, we have new guidelines but haven't published new `contributing.md` yet, so I hope that small update is okay.

dmueller:
> Great detective work @dmueller and @gleonard-m for catching this issue!
> 
> I feel like this isn't the first time we've had a situation where preview wasn't correctly calling UAPI and was obscuring important info, and I'm wondering if we can catch this more proactively somehow ...

I agree with this but will have to mull on it. At the surface level, if it is showing ads then it *looks* like it is working which hides it the most. One step back from there, the unified preview path using the same dropdowns when support isn't in parity yet also causes confusion (e.g. the country/region overrides still do nothing on that one); and then one more step back the "debug info" could be improved (capture the API request(s) it made and show them, etc.) which should help it be realized quicker.

